PR #3, opened 2020-11-02
Fix CI_TOKEN for dev-scripts

abays:
Ran into issues doing a fresh install on the new machines.  This was required to get dev-scripts to work.


stuggi:
we can re-introduce this, I basically changed it that you add your CI token to the local-defaults file instead of a separate file. The local-defaults file then looks like:

~~~
secrets_repo: https://...
ci_token: YOURTOKEN
~~~

abays:
That works too.  Let me just change this to a check to make sure `ci_token` is set and is not the empty string.  It could of course still fail later if you have a garbage string, but at least it will catch it sooner for the case when you just forget to set it.

PR #15, opened 2020-12-04
Deploy overcloud using tripleo_deploy from openstackclient pod

stuggi:
This prepares the configmaps use to deploy the overcloud via the
openstackclient pod.

Also adds Dockerfile for the openstackclient container image.


ansible/files/osp/tripleo_deploy/deployed-server-port-map.yaml
```diff
@@ -0,0 +1,40 @@
...
+  OS::TripleO::Network::Ports::OVNDBsVipPort: /usr/share/openstack-tripleo-heat-templates/network/ports/noop.yaml
+
+parameter_defaults:
+  DeployedServerPortMap:
+    control_virtual_ip:
+      fixed_ips:
+        - ip_address: 192.168.25.100
+      subnets:
+        - cidr: 192.168.25.0/24
+      network:
```
dprince, line 14: What does the network tags bit do here? I wasn't aware we needed that.

dprince:
lgtm on the patch. An inline question I mentioned above.

ansible/files/osp/tripleo_deploy/deployed-server-port-map.yaml
stuggi: this is only needed for the ctrlplane network: https://github.com/openstack/tripleo-heat-templates/blob/master/overcloud.j2.yaml#L553-L555

From https://github.com/openstack/tripleo-heat-templates/blob/master/releasenotes/notes/make-NetCidrMapValue-contain-lists-of-cidrs-19bc98dc9bc6b819.yaml : 
```
---
upgrade:
  - |
    Tags are now used on the ``ctlplane`` network to store the list of cidrs
    associated with the subnets on the ``ctlplane`` network. Users of
    Deployed Server (pre-provisioned servers) need to update the port map
    (``DeployedServerPortMap``) to include the required data. For example::

      parameter_defaults:
        DeployedServerPortMap:
          controller0-ctlplane:
            fixed_ips:
              - ip_address: 192.168.24.9
            subnets:
              - cidr: 192.168.24.0/24
            network:
              tags:
                - 192.168.24.0/24
                - 192.168.25.0/24
          compute0-ctlplane:
            fixed_ips:
              - ip_address: 192.168.25.8
            subnets:
              - cidr: 192.168.25.0/24
            network:
              tags:
                - 192.168.24.0/24
                - 192.168.25.0/24
```

PR #21, opened 2021-01-04
Disable existing repo files

stuggi:
If e.g. epel is enabled from a previous usage of the host, dev-scripts
install fails. Lets disable all existing repository definitions as we
register to what is required.

### DIFF
--- a/ansible/host_prep.yaml
+++ b/ansible/host_prep.yaml
@@ -7,19 +7,20 @@
   - name: Include variables
     include_vars: vars/default.yaml
 
-  - name: Remove beaker repos
+  - name: Disable existing repos
     block:
-    - name: get beaker repo files
+    - name: get existing repo files
       find:
         paths: /etc/yum.repos.d
-        patterns: beaker*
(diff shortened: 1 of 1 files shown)


ansible/host_prep.yaml
```diff
@@ -7,19 +7,19 @@
...
-        patterns: beaker*
-      register: beaker_files_to_delete
+        patterns: '*.repo'
+      register: repo_files_to_delete
 
-    - name: Delete beaker repo files
+    - name: Delete existing repo files
       file:
         path: "{{ item.path }}"
         state: absent
```
mpryc, line 21: I am unsure, but maybe there should be some backup made before deletion, or disabling instead of deleting? e.g.: https://github.com/redhat-openstack/octario/blob/master/roles/setup_repos/tasks/main.yml#L2-L16

stuggi: yes I'll disable them instead of deleting.

stuggi: done

PR #25, opened 2021-01-05
Add Makefile target to create network(s)

dprince:
Currently creates just the overcloud ctlplane. Additionally
this patch is currently only integrated with baremetalset.

Working on integration with the ctlplane VM next...


stuggi (approved):
lgtm

PR #34, opened 2021-01-13
Migrate template files into osp-director-operator

stuggi:
We move manadory deployment templates into the osp-director-operator.
Thos can be removed here.


ansible/host_prep.yaml
```diff
@@ -38,25 +38,6 @@
...
-          priority=2
-          includepkgs=redhat-internal-cert-install
-        mode: '0644'
-
-    - name: Install helpful packages
-      package:
-        state: installed
-        name:
-          - redhat-internal-cert-install
-
```
stuggi, line 59: use GIT_SSL_NO_VERIFY and removed internal repo

dprince (approved):

PR #43, opened 2021-01-19
Support install/cleanup of Local Storage Operator

dprince:
This adds configurable support for the Local Storage Operator
within our dev-scripts. It automatically attaches disks to
the configured worker VMs and configures the local-storage-operator
to provide those as PVs for the 'local' storage class.

Two Makefile targets are included:

 -make local_storage
 -make local_storage_cleanup

These targets are both currently optional (opt in) for those who
wish to experiment with an alternate PV type.

### DIFF
--- a/ansible/local_storage.yaml
+++ b/ansible/local_storage.yaml
@@ -107,7 +107,7 @@
     - "storage_class.yaml"
     - "local_disks.yaml"
 
-  - name: Create namespace, operatorgroup and subscription
+  - name: Creating storage class and local disks
     shell: |
       set -e
       oc apply -f "{{ local_storage_volume_yaml_dir }}"
--- a/ansible/local_storage.yaml
+++ b/ansible/local_storage.yaml
@@ -1,0 +1,115 @@
+---
+- hosts: localhost
+  vars_files: vars/default.yaml
+  become: true
+  become_user: root
+  roles:
+  - oc_local
+
+  tasks:
+  - name: Create VM attached Local Storage for local-storage-operator
+    with_items: "{{ local_storage_domains }}"
+    include_role:
+      name: local_storage
+    vars:
+      domain: "{{ item }}"
+
+  - name: Lookup defaultChannel for local-storage-operator
+    shell: >
+      oc get -o json packagemanifest local-storage-operator | jq -re .status.defaultChannel
+    register: local_storage_default_channel_cmd
+    environment: &oc_env
+      PATH: "{{ oc_env_path }}"
+      KUBECONFIG: "{{ kubeconfig }}"
+
+  - name: Set defaultChannel fact for local-storage-operator
+    set_fact:
+      local_storage_default_channel: "{{ local_storage_default_channel_cmd.stdout }}"
+
+  - name: Lookup currentCSV for local-storage-operator
+    shell: >
+      oc get -o json packagemanifest local-storage-operator | jq -re .status.channels[0].currentCSV
+    register: local_storage_current_csv_cmd
+    environment:
+      <<: *oc_env
+
+  - name: Set currentCsv fact for local-storage-operator
+    set_fact:
+      local_storage_current_csv: "{{ local_storage_current_csv_cmd.stdout }}"
+
+  - name: Set directory for local_storage yaml files
+    set_fact:
+      local_storage_yaml_dir: "{{ working_yamls_dir }}/local_storage_operator"
+
+  - debug:
+      msg: "Operator yamls will be written to {{ local_storage_yaml_dir }} locally"
+
+  - name: Create local yamldir for local_storage operator
+    file:
+      path: "{{ local_storage_yaml_dir }}"
+      state: directory
+      mode: '0755'
+      owner: root
+      group: root
+
+  - name: Render templates to local_storage operator yaml dir
+    template:
+      src: "local_storage/operator/{{ item }}.j2"
+      dest: "{{ local_storage_yaml_dir }}/{{ item }}"
+      mode: '0644'
+      owner: root
+      group: root
+    with_items:
+    - "namespace.yaml"
+    - "operatorgroup.yaml"
+    - "subscription.yaml"
+
+  - name: Create namespace, operatorgroup and subscription for Local Storage Operator
+    shell: |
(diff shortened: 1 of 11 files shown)


mpryc:
Please rebase this PR, it should allow downstream CI to pass.

ansible/roles/local_storage/tasks/cleanup.yaml
```diff
@@ -0,0 +1,45 @@
...
+          if [ -f "$fs" ]; then
+              rm -rf "{{ local_storage_data_dir }}/{{ domain }}"
+          fi
+
+      done
+
+  - name: Stop {{ domain }} VM
+    virt:
+      name: "{{ domain }}"
+      state: destroyed
```
abays, line 44: Do we want to destroy the VM here?  Just wondering if that should be left to our current mechanism for VM clean-up, since this is just meant to clean local storage disks

ansible/local_storage.yaml
```diff
@@ -0,0 +1,106 @@
...
+      oc apply -f "{{ local_storage_yaml_dir }}"
+    environment:
+      <<: *oc_env
+
+  - name: wait for OpenShift Local Storage Operator to be installed
+    shell: |
+      oc wait deployment.apps/local-storage-operator -n openshift-local-storage --for condition=Available --timeout={{ default_timeout }}s
+    environment:
+      <<: *oc_env
+    retries: 50
```
mpryc, line 74: default_timeout is 240 which /5 is 48, maybe this should be more precise and for example:

retries: '{{ (default_timeout / 5) | int }}'

ansible/roles/local_storage/tasks/cleanup.yaml
```diff
@@ -0,0 +1,45 @@
+---
+- name: Get current {{ domain }} VM attached disk count
+  become: true
+  become_user: root
+  shell: "virsh dumpxml {{ domain }} | grep \"source file='{{ local_storage_data_dir }}/{{ domain }}/local_storage_disk\" | wc -l"
+  register: cur_storage_disks
+
+- name: Change {{ domain }} VM specs
+  when: (cur_storage_disks.stdout | int) > 0
+  block:
```
mpryc, line 10: I see that only virt task do not require root access, is this correct?
Maybe become/become_user should be moved to be applied for entire block

ansible/roles/local_storage/tasks/cleanup.yaml
```diff
@@ -0,0 +1,45 @@
...
+    ignore_errors: yes
+    loop: "{{ local_storage_disks }}"
+    loop_control:
+      index_var: index
+      loop_var: disk
+
+  - name: Destroy local storage data disks for {{ domain }} VM
+    become: true
+    become_user: root
+    shell: |
```
mpryc, line 29: I wonder if ansible file module would not handle that? the following removes /tmp/a with everything inside this dir, so loop over {{ local_storage_disks }} should be enough ?

    - file:
        state: absent
        path: "/tmp/a/mydir/subdir"

ansible/roles/local_storage/tasks/cleanup.yaml
dprince: The VM isn't 'destroyed' as in 'virsh undefined'. But we are 'virsh destroy'ing the VM to stop it immediately. It will get restarted automatically... thus cleanup should work.

ansible/local_storage.yaml
```diff
@@ -0,0 +1,106 @@
...
+      local_storage_yaml_dir: "{{ working_yamls_dir }}/local_storage_operator"
+
+  - debug:
+      msg: "Operator yamls will be written to {{ local_storage_yaml_dir }} locally"
+
+  - name: Create local yamldir for local_storage operator
+    file:
+      path: "{{ local_storage_yaml_dir }}"
+      state: directory
+      mode: '0755'
```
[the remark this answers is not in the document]
dprince: I can set this, but fwiw this is the same snippet we use everywhere in this project. We should change them all.

ansible/local_storage_cleanup.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+  - name: cleanup local pvs
+    command: "oc delete pv {{ item }}"
+    environment:
+      <<: *oc_env
+    ignore_errors: true
+    with_items: "{{ local_pvs.stdout_lines | list }}"
+
+  - name: Cleanup VM attached Local Storage
+    with_items: "{{ local_storage_domains }}"
```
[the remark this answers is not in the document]
dprince: done.

ansible/local_storage_cleanup.yaml
```diff
@@ -0,0 +1,44 @@
+---
+- hosts: localhost
+  vars_files: vars/default.yaml
+  user: root
```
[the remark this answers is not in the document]
dprince: sure.

ansible/local_storage.yaml
dprince: I find the suggested logic actually more difficult to understand. To me 'retries' is just retries... not seconds.

ansible/roles/local_storage/tasks/cleanup.yaml
dprince: got it in the latest patch.

ansible/roles/local_storage/tasks/cleanup.yaml
dprince: done.

ansible/local_storage.yaml
```diff
@@ -0,0 +1,115 @@
...
+      src: "local_storage/volume/{{ item }}.j2"
+      dest: "{{ local_storage_volume_yaml_dir }}/{{ item }}"
+      mode: '0644'
+      owner: root
+      group: root
+    with_items:
+    - "storage_class.yaml"
+    - "local_disks.yaml"
+
+  - name: Create namespace, operatorgroup and subscription
```
abays, line 110: Maybe this should say "create storage class and local disks"?  I might be misreading something though

dprince: got it. thanks, this was a remnant copied from above I think

PR #47, opened 2021-01-25
Temporary webhook deployment support

abays:
OLM seems to be incapable of completely deploying all required resources for admission controller validation webhooks to function properly.  For the meantime, we can use dev-tools to deploy the two resources (the actual webhook YAML and accompanying service) that OLM fails to bundle.

NOTE: If we decide to merge this, it must *not* be merged before https://github.com/openstack-k8s-operators/osp-director-operator/pull/110 is merged and a new `osp-director-operator:0.0.1` is built from master and pushed to `quay.io/openstack-k8s-operators`.


abays:
Opting-out of this approach

PR #48, opened 2021-01-27
Add datavolume playbook

dprince:
This playbook pre-creates an openstack-base-img DataVolume
using virtctl on the CLI. This has the effect of uploading
the image into the cluster from disk thus avoiding
repeatedly downloading the same image. This also works
around a case where cdi-datavolume streamed http downloads
can be really slow from some mirrors (more to be said about
that but perhaps not here).

### DIFF
--- a/ansible/datavolume.yaml
+++ b/ansible/datavolume.yaml
@@ -1,0 +1,45 @@
+#!/usr/bin/env ansible-playbook
+---
+- hosts: localhost
+  vars_files: "vars/default.yaml"
+  roles:
+  - oc_local
+
+  tasks:
+
+  # NOTE: we copy this to the Ironic images directory to reuse it for
+  # the provision server (baremetalsets). This avoids downloading
+  # the same image twice.
+  - name: Check if {{ osp_controller_base_image_url | basename }} already exist
+    stat:
+      path: "{{ base_path }}/ironic/html/images/{{ osp_controller_base_image_url | basename }}"
+    register: stat_result
+
+  - name: Get RHEL guest base image
+    when: not stat_result.stat.exists
+    get_url:
+      url: "{{ osp_controller_base_image_url }}"
+      dest: "{{ base_path }}/ironic/html/images/{{ osp_controller_base_image_url | basename }}"
+      mode: '0644'
+
+  - name: does the datavolume already exist
+    ignore_errors: true
+    shell: >
+      oc get datavolume openstack-base-img -n {{ namespace }}
+    environment: &oc_env
+      PATH: "{{ oc_env_path }}"
+      KUBECONFIG: "{{ kubeconfig }}"
+    register: datavolume_switch
+
+  - name: Deploy the datavolume
+    shell: |
+      virtctl image-upload dv openstack-base-img -n {{ namespace }} --size={{ osp_controller_disk_size }}G --image-path=/home/ocp/ironic/html/images/{{ osp_controller_base_image_url | basename }}
+    environment:
+      <<: *oc_env
+    when: datavolume_switch.rc == 1
+
+  - name: Wait for the datavolume to be ready
+    shell: |
+      oc wait datavolume openstack-base-img --for condition=Ready -n "{{ namespace }}" --timeout={{ default_timeout }}s
+    environment:
+      <<: *oc_env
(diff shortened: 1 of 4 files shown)


ansible/install_computes.yaml
```diff
@@ -26,18 +26,6 @@
       PATH: "{{ oc_env_path }}"
       KUBECONFIG: "{{ kubeconfig }}"
 
-  - name: Check if {{ osp_controller_base_image_url | basename }} already exist
```
stuggi, line 29: maybe we do to not remove this and make it only run if osp_controller_base_image_url is set? with this we could then cover automation testing for both cases.

ansible/datavolume.yaml
```diff
@@ -0,0 +1,42 @@
...
+- hosts: localhost
+  vars_files: "vars/default.yaml"
+  roles:
+  - oc_local
+
+  tasks:
+
+  - name: Check if {{ osp_controller_base_image_url | basename }} already exist
+    stat:
+      path: "/home/ocp/ironic/html/images/{{ osp_controller_base_image_url | basename }}"
```
mpryc, line 12: base_path is set in ansible/vars/default.yaml, so:

"{{ base_path }}/ironic/html/images/{{ osp_controller_base_image_url | basename }}"

ansible/datavolume.yaml
```diff
@@ -0,0 +1,42 @@
...
+    register: stat_result
+
+  - name: Get RHEL guest base image
+    when: not stat_result.stat.exists
+    get_url:
+      url: "{{ osp_controller_base_image_url }}"
+      dest: "/home/ocp/ironic/html/images/{{ osp_controller_base_image_url | basename }}"
```
mpryc, line 19: Same here for base_path

ansible/install_computes.yaml
abays: Yeah, we could wrap this in a `block` with a `when` that checks `osp_controller_base_image_url`

ansible/datavolume.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+  tasks:
+
+  - name: Check if {{ osp_controller_base_image_url | basename }} already exist
+    stat:
+      path: "/home/ocp/ironic/html/images/{{ osp_controller_base_image_url | basename }}"
+    register: stat_result
+
+  - name: Get RHEL guest base image
+    when: not stat_result.stat.exists
```
mpryc, line 16: Is there a way to get the checksum of the RHEL guest base image?
If yes then maybe it's better not to check stat_result.stat.exists and instead write as:

    get_url:
      url: "{{ osp_controller_base_image_url }}"
      dest: "{{ base_path }}/ironic/html/images/{{ osp_controller_base_image_url | basename }}"
      checksum: "sha256:{{ ocp_controller_base_image_md5sum }}"

This way ansible will take care of the image being already downloaded and on top of that the validity of the image will be checked.

ansible/install_computes.yaml
dprince: I think I can accommodate the request here but do note that osp_controller_base_image_url is not a new variable added in this patch. I moved this part over from install_computes.yaml to deduplicate the logic and only download the image in one place.

It sounds like what you guys want is a dedicated parameter to enable/disable the use of the image URL on the CRD right?

stuggi: sorry I missed somehow that it is moved. I think we are good atm. we can do add how we test the different scenarios later

ansible/datavolume.yaml
dprince: got it fixed in the latest patch

ansible/datavolume.yaml
dprince: got it fixed in the latest patch

ansible/datavolume.yaml
dprince: I don't see the checksum's on the download server we use for these so I'm not sure this is possible.

Perhaps we follow up on this as I didn't write this in this patch... just moved it over from install_computes.yaml to avoid duplication.

ansible/install_computes.yaml
dprince: we spoke about this on slack and I think the functionality here can come as a follow up

ansible/datavolume.yaml
mpryc: Yep there are none.

PR #50, opened 2021-01-29
Reduce VM disk size for masters and workers

mpryc:
This reduces disk size for VMs to the level which should be
sufficient for standard deployment. Usually master nodes
were around 15G each, similarly for worker nodes.

### DIFF
--- a/ansible/vars/default.yaml
+++ b/ansible/vars/default.yaml
@@ -31,10 +31,10 @@ ocp_num_extra_workers: 2
 ocp_extra_workers_online_status: false
 ocp_master_memory: 16384
 ocp_master_vcpu: 8
-ocp_master_disk: 120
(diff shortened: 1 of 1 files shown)


ansible/vars/default.yaml
```diff
@@ -31,10 +31,10 @@ ocp_num_extra_workers: 2
 ocp_extra_workers_online_status: false
 ocp_master_memory: 16384
 ocp_master_vcpu: 8
-ocp_master_disk: 120
+ocp_master_disk: 50
```
stuggi, line 34: we probably want to switch it. 50GB for the workers where we run the controller VMs nested and 30GB for the masters.

stuggi (approved):
/lgtm

PR #51, opened 2021-01-29
Use ocp user when downloading RHEL guest image

mpryc:
RHEL guest image is downloaded into ironic subfolder
owned by the ocp user, and downloading of RHEL guest
image must be perfomed by the same user to succeed.

Changing also from /home/ocp to the variable base_path
defined in defaults.yaml.


stuggi (approved):
/lgtm

stuggi (approved):
/lgtm

PR #54, opened 2021-02-01
Add demo playbook

dprince:
This creates the demo user and stages OLM for a demo

### DIFF
--- a/ansible/demo.yaml
+++ b/ansible/demo.yaml
@@ -6,11 +6,6 @@
 
   tasks:
 
-  #- name: install python3-passlib if needed
-  #  package:
-  #    name: python3-passlib
-  #    state: present
-
   - name: create demo directory
     file:
       path: "{{ working_dir }}/demo"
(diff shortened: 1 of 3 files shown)


ansible/demo.yaml
```diff
@@ -0,0 +1,80 @@
+---
+- hosts: localhost
+  vars_files: "vars/default.yaml"
+  roles:
+  - oc_local
+
+  tasks:
+
+  #- name: install python3-passlib if needed
```
dprince, line 9: Didn't see this package so I 'pip installed' it locally. Perhaps we just pip install it here?

mpryc: How about using simple template file or even lineinfile ansible module? This way we won't need the python*-passlib dependency for ansible module. I would say yes for such dependency if it's more complex then just demo/demo password. Note this is MD5 sum, Bcrypt is better one when using hashes for passwords e.g.
In Bcrypt the line will  be:
 demo:$2y$10$WrEag9RyuNF/294Z7QmtUeSxLiHATgtbG5.lOBonc0Xm.hRmyCzLq


    - name: create htpasswd file
      lineinfile:
        path: "{{ working_dir }}/demo/htpasswd"
        line: 'demo:$apr1$SeAzqzN/$UPXamQYHPy/mlyIxoyQ9b0'
        state: present
        create: yes

mpryc: There is yet another way to skip the dependency from passlib, it's not 1000 times iterated MD5 sum as Apache auth, but should do the job:

- name: create htpasswd file
  lineinfile:
    path: /tmp/htpasswda
    line: 'demo:{{ "demo" | password_hash("md5") }}'
    state: present
    create: yes

abays: Whatever you guys want to do here is fine with me, but otherwise this all looks good

dprince: ack, lets use this approach

stuggi (approved):
/lgtm

PR #55, opened 2021-02-02
Bump to OCP 4.6.13 and CNV 2.5.3.

dprince:


ansible/vars/default.yaml
```diff
@@ -5,7 +5,7 @@ base_path: /home/ocp
 
 # To set a specific release to install.
 ocp_version: 4.6
-ocp_release_image: quay.io/openshift-release-dev/ocp-release:4.6.12-x86_64
+ocp_release_image: quay.io/openshift-release-dev/ocp-release:4.6.13-x86_64
```
stuggi, line 8: there is also -15. should we bump to that latest?

dprince: I wonder if its not stable yet. I'm not seeing that in the stable channel in my OCP clusters yet at least. Perhaps lets just hold off for now as .13 seems to work okay.

PR #56, opened 2021-02-02
Align OLM cleanup for webUI and CLI

dprince:
The naming schemes for the subscription and operator group
were slightly different. This patch aligns things so that
olm_cleanup will remove the installed OLM resources regardless
of whether the CLI or UI is used to install them.

The operatorgroup resource gets auto-created with a generated
name via the UI so we add logic to the playbook to lookup and use
that for deletion accordingly.

### DIFF
--- a/ansible/demo.yaml
+++ b/ansible/demo.yaml
@@ -54,13 +54,13 @@
 
   - name: delete OLM subscription
     shell: |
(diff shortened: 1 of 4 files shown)


abays (approved):
/lgtm

PR #61, opened 2021-02-11
Improve logic in install_networks

dprince:
Better implementation that doesn't show as red
if the overcloud net doesn't exist yet (on initial creation)


ansible/install_networks.yaml
```diff
@@ -31,19 +31,13 @@
...
-    environment: &oc_env
-      PATH: "{{ oc_env_path }}"
-      KUBECONFIG: "{{ kubeconfig }}"
-    register: openstack_net_switch
-
   - name: Start networks
     shell: |
       set -e
-      oc apply -n openstack -f "{{ network_yaml_dir }}"
+      if [ -z "$(oc get osnet ctlplane -o json --ignore-not-found)" ]; then
```
stuggi, line 37: we might want to add -n openstack

dprince: got it in latest rebase.

ansible/install_networks.yaml
```diff
@@ -31,19 +31,21 @@
     with_items:
     - "01-ctlplane.yaml"
 
-  - name: do the networks already exist
+
+  - name: does the provisionserver already exist
```
stuggi, line 35: provisionserver -> network

dprince: got it in latest commit. Thanks.

stuggi (approved):
/lgtm

PR #62, opened 2021-02-11
Create shared provisionserver

dprince:
This uses the new ProvisionServerName feature on BaremetalSet
to leverage use of a pre-created ProvisionServer when creating
the CR. Should speed up subsequent deployments of BaremetalSets.

### DIFF
--- a/ansible/install_computes.yaml
+++ b/ansible/install_computes.yaml
@@ -37,11 +37,28 @@
       dest: "{{ compute_yaml_dir }}/{{ item }}"
       mode: '0644'
     with_items:
+    - "openstackprovisionserver.yaml"
     - "openstackbaremetalset.yaml"
 
-  - name: Start openstackbaremetalset
+  - name: does the provisionserver already exist
+    ignore_errors: true
+    shell: >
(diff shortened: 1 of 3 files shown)


dprince:
This won't pass until we land: https://github.com/openstack-k8s-operators/osp-director-operator/pull/147

ansible/install_computes.yaml
```diff
@@ -37,11 +37,28 @@
...
       mode: '0644'
     with_items:
+    - "openstackprovisionserver.yaml"
     - "openstackbaremetalset.yaml"
 
-  - name: Start openstackbaremetalset
+  - name: does the provisionserver already exist
+    ignore_errors: true
+    shell: >
+      oc get osprovserver openstack -o json --ignore-not-found
```
stuggi, line 46: should add -n openstack

stuggi (changes requested):

ansible/install_computes.yaml
dprince: got it in latest commit, thanks!

stuggi (approved):
/lgtm

PR #66, opened 2021-02-18
Rename controlplane and baremetalset

dprince:


abays (approved):
Aside from the checks having issues, looks good to me

PR #68, opened 2021-02-22
Pin ansible-lint version to the latest 4.x

mpryc:
Ansible-lint introduced number of changes in 5.x which
gives errors until the playbooks. Until playbooks are fixed
to work with 5.x we need to pin to use 4.x.

See:
  https://github.com/ansible-community/ansible-lint/discussions/1150


stuggi (approved):
/lgtm

PR #78, opened 2021-03-09
SRIOV operator install

abays:
Playbook and Makefile logic to install OpenShift SRIOV Network Operator


ansible/files/sriov/subscription.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: sriov-network-operator-subsription
+  namespace: openshift-sriov-network-operator
+spec:
+  channel: "4.6"
```
stuggi, line 7: maybe this can be a var or if it always matches the ocp version, use ocp_version

stuggi (approved):
/lgtm

PR #79, opened 2021-03-12
Upgrade pip version before installing oc client

jlarriba:
If pip is not the last version, it fails to install the necesary cryptography dependency.

### DIFF
--- a/ansible/local_oc_client.yaml
+++ b/ansible/local_oc_client.yaml
@@ -12,6 +12,10 @@
       - python3-pip
     become: yes
 
+  - name: Upgrade pip
+    shell: |
+      pip3 install --upgrade pip
+
   - name: Install openstack client
     shell: |
       pip3 install --user python-openstackclient osc-placement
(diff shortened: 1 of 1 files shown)


ansible/local_oc_client.yaml
```diff
@@ -12,6 +12,10 @@
       - python3-pip
     become: yes
 
+  - name: Upgrade pip
+    shell: |
+      pip3 install --upgrade pip
+
```
abays, line 18: I wonder if we need `become: true` here, because this is affecting the `pip3` binary itself.  When I try to run the command manually as a non-root user, I get this error:
```
PermissionError: [Errno 13] Permission denied: '/usr/local/lib/python3.6/site-packages/pip'
```

jlarriba: You are right. I was thinking about it. I launched it without become, but I was the root user...

Added!

abays:
Thanks @jlarriba!

/lgtm

abays (approved):

PR #81, opened 2021-03-15
rhel-subscription does not need to be mandatory

jlarriba:
This PR does not need to go through if you think it is an option you do not want to give, but it is pretty useful for me, so I thought I might help others too. Feel free to dismiss if it is not something you would like.

Currently, passing a rhel subscription key and orgid are required. If you are using RHEL8 to run the playbook (which is necessary due to the use of rhel-8 repositories), the machine should have already been subscribed in order to download the ansible executable. As long as most of the subscriptions currently give access to the two needed repos, it is not needed to unsubscribe and resuscribe, just to import the repos.

So, I made a little modification in which, if the rhel-subscription.yaml file is not found, we just log a message indicating that the current subscription will be used by the script to add the repos.

The main motivator for this change (although not the unique) is that it does not seem possible (or at least I did not find anywhere or anywho who knows how) to obtain an apikey and an orgid from a "developer" subscription, which is mainly what the people running this playbooks will be using.

### DIFF
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -31,14 +31,23 @@ dnf install -y ansible git libvirt-client python3-netaddr python3-lxml
 
 First create the following files relative to the project root:
 
-`vars/rhel-subscription.yaml` in the format:
+`files/pull-secret` which contains your OCP pull-secrets file.
+
+`vars/rhel-subscription.yaml` is optional. File format:
 
 ```
 rhel_subscription_activation_key: xyz
 rhel_subscription_org_id: "123123123"
 ```
 
-`files/pull-secret` which contains your OCP pull-secrets file.
+If `vars/rhel-subscription.yaml` is not specified, a subscription must have been configured manually prior to running the ansible. The manual subscription must give access to the following repositories:
+
+```
+advanced-virt-for-rhel-8-x86_64-rpms                    Advanced Virtualization for RHEL 8 x86_64 (RPMs)
+ansible-2-for-rhel-8-x86_64-rpms                        Red Hat Ansible Engine 2 for RHEL 8 x86_64 (RPMs)
+rhel-8-for-x86_64-appstream-rpms                        Red Hat Enterprise Linux 8 for x86_64 - AppStream (RPMs)
+rhel-8-for-x86_64-baseos-rpms                           Red Hat Enterprise Linux 8 for x86_64 - BaseOS (RPMs)
+```
 
 Any other parameter from `vars/default.yaml` can be customized using
 `local-defaults.yaml`. Add parameters to meet your req.
--- a/ansible/host_prep.yaml
+++ b/ansible/host_prep.yaml
@@ -30,13 +30,19 @@
       - name: Include rhel-subscription info
         include_vars: vars/rhel-subscription.yaml
       rescue:
-      - fail:
-          msg: |
-            vars/rhel-subscription.yaml is not present. You must create this
+      - name: No rhel-subscription info
+        pause:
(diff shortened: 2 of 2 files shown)


ansible/README.md
```diff
@@ -54,7 +54,7 @@ export CI_TOKEN: <TOKEN>
 Additionally create the following files relative to the
 project root:
 
-`vars/rhel-subscription.yaml` in the format:
+`vars/rhel-subscription.yaml` is optional. If the file is not created, it will try to use your current subscription to add the corresponding repositories:
```
stuggi, line 57: we probably should then list the repose which are required to install the required packages:

~~~
advanced-virt-for-rhel-8-x86_64-rpms                                                             Advanced Virtualization for RHEL 8 x86_64 (RPMs)
ansible-2-for-rhel-8-x86_64-rpms                                                                 Red Hat Ansible Engine 2 for RHEL 8 x86_64 (RPMs)
rhel-8-for-x86_64-appstream-rpms                                                                 Red Hat Enterprise Linux 8 for x86_64 - AppStream (RPMs)
rhel-8-for-x86_64-baseos-rpms                                                                    Red Hat Enterprise Linux 8 for x86_64 - BaseOS (RPMs)
~~~

ansible/host_prep.yaml
```diff
@@ -58,27 +63,34 @@
...
         pool: '^(Red Hat Enterprise Server for x86_64)$'
+      when: manual_rhel_subscription is undefined
 
     - name: Register with activationkey and consume subscriptions matching Red Hat Enterprise Server
       redhat_subscription:
         state: present
         activationkey: "{{ rhel_subscription_activation_key }}"
         org_id: "{{ rhel_subscription_org_id }}"
         pool: '^(Red Hat Enterprise Server for x86_64)$'
+      when: manual_rhel_subscription is undefined
```
stuggi, line 93: we can move this into a block and only run both tasks `when: manual_rhel_subscription is undefined`

ansible/host_prep.yaml
```diff
@@ -58,27 +63,34 @@
...
             rhel-subscription.yaml is not present in {{ secrets_repo }}. You must create this
             file. The format of the file is:
 
             rhel_subscription_activation_key: <activation key>
             rhel_subscription_org_id: "xxxxxxx"
 
+            If you do not have the file, we will use your manually provisioned
+            subscription
+        register: manual_rhel_subscription
+
```
stuggi, line 78: wouldn't it be enough to extend the condition of the "use secret_repo" block to check manual_rhel_subscription ?

stuggi (changes requested):

ansible/host_prep.yaml
jlarriba: Im sorry, but I do not understand what you are proposing here. Could you give a little bit more context of what you see wrong?

stuggi: sorry, after reading again today on what I wrote yesterday, my comment does not make sense :)

stuggi (approved):
/lgtm

akiselev1:
I had exactly the same problem, and @abays pointed me to this PR.  It was not merged back then for some reason.
@jlarriba - let me know if you are still interested in this change. I can re-test it with the current code and resolve any merge conflicts if needed. Just give me access to your fork/branch.

jlarriba:
Hi @akiselev1 , of course I am still interested!

I think that running the ansible playbook on an already subscribed RHEL machine is a pretty common use case.

abays (approved):
/lgtm

PR #82, opened 2021-03-15
Virtualized SRIOV install option

abays:
Provides the option to turn a certain number of OCP workers into VMs capable of supporting virtualized SRIOV.  Best used when installing a cluster, but can also work after-the-fact.


ansible/virt_sriov.yaml
```diff
@@ -0,0 +1,170 @@
...
+        copy:
+          src: /usr/share/qemu-kvm
+          dest: /usr/share/qemu
+          mode: 0755
+          remote_src: yes
+      when: not custom_qemu.stat.exists
+    
+    - name: Set directory for Libvirt XML files
+      set_fact:
+        xml_dir: "/tmp/sriov/xml"
```
stuggi, line 102: we could also put this in {{ working_dir }}/sriov where all the other files are located

abays: I was wondering about this.  The thing is, it looks like the `working_dir` is typically used with `hosts: localhost` playbooks.  In this case, the `virsh` binaries and libvirt daemon are on the target provisioning machine.  That's why I put the XMLs on the `convergence_base` host (which I know is often the same thing as the `localhost`, but isn't necessarily).  But I might be misunderstanding where the `working_dir` actually lives.

dprince (approved):

PR #85, opened 2021-03-22
Add isolated network support

stuggi:
Introduce osp_network_type varibale which allows to control
the deployed network architecture, ATM either flat or vlan.
Default to vlan.

Creates the required different osnets for network isolation
by separate each network (except ctlpane) into its own vlan.

Creates external network vlan and gateway interface via
libvirt network hooks script.

Depends-On: https://github.com/openstack-k8s-operators/osp-director-operator/pull/184


ansible/files/qemu/network
```diff
@@ -0,0 +1,52 @@
...
+    ip addr add 10.0.0.1/24 dev ospnetwork.10
+    if [ $? -eq 0 ]; then
+      logger -p local3.info "qemu hook script - external vlan interface ospnetwork.10 created successfully"
+    else
+      echo "failed to create vlan interface ospnetwork.10" >&2
+      exit 1
+    fi
+
+    ip link set ospnetwork.10 up
+    ping -c 1 -w 5 10.0.0.1 >/dev/null 2>&1
```
abays, line 23: What interface is used for this ping test?  Just wanted to make sure it doesn't use the same device (`ospnetwork.10`) to ping itself.

abays (commented):
Looks good to me, though someone with more OOO experience should review those particular templates

ansible/files/qemu/network
stuggi: it will use the same interface it is the only one in the vlan on the host. the test is just to make sure the ip was correct assigned. Should be ok, no?

abays: Ah, okay.  That makes sense then

dprince (approved):

PR #91, opened 2021-03-31
Bump to OCP 4.7, CNV 2.6

dprince:
The check for osp-director-operator-index was removed as this
pod is now named with a unique ID appended. Also the check isn't
explicitly needed as the packagemanifest check below is really what matters.


stuggi (approved):
/lgtm

PR #95, opened 2021-04-20
Add ComputeHCI support

stuggi:
deploying Ceph with ComputeHCI can now be controlled with osp_compute_hci: true
Before creating the osbms the shutoff worker VMs get additional OSD disks created
and attached.

Depends-On: https://github.com/openstack-k8s-operators/osp-director-operator/pull/210


ansible/install_computes.yaml
```diff
@@ -23,6 +23,29 @@
       state: directory
       mode: '0755'
 
+  - name: Detach additional Ceph OSD disks
+    shell: |
+      #!/bin/bash
+      for domain in $(virsh list --inactive --name); do
```
abays, line 29: Does this always return just the stopped workers?  Just curious if we might need a little more specificity here just in case other VMs are sitting around powered-off.

abays (commented):
Looks fine as far as I can tell.  Probably want Dan to take a look too though

ansible/install_computes.yaml
stuggi: no, it'll return all stopped VMs. I think in general there should be only stopped OCP workers when this playbook runs, but we can limit the output for the VMs

stuggi: just a note, there is an active PR worked on to get additional features in dev-scripts. so these steps then could be removed

PR #97, opened 2021-04-27
Remove cleanup before dev-scripts deploy

stuggi:
With enabling assisted installer in dev-scripts, cleanup job fail
if connection to the ocp cluster is not possible using oc client.

For now disable dev-scripts cleanup run.


abays (approved):
/lgtm

PR #98, opened 2021-04-27
Assisted installer deploy option

abays:
Behemoth patch that enables our dev-tools to deploy an OCP cluster via the new-ish assisted installer.  Current restrictions on usage:

1. Provisioning host must be RHEL 8.3
2. Only version 4.7 of OCP is supported
3. Netmasks for CIDRs must be `/24`
4. IPV6 not supported

Basic usage: just set `ocp_ai: true` in your `local-defaults.yaml`

### DIFF
--- a/ansible/hosts
+++ b/ansible/hosts
@@ -1,3 +1,6 @@
+[all:vars]
+ocp_ai=False
+
 localhost ansible_connection=local
 
 [convergence_base]
--- a/ansible/hosts
+++ b/ansible/hosts
@@ -1,7 +1,7 @@
-[all:vars]
-ocp_ai=False
-
 localhost ansible_connection=local
 
 [convergence_base]
 localhost
+
+[all:vars]
+ocp_ai=False
--- a/ansible/install_networks.yaml
+++ b/ansible/install_networks.yaml
@@ -23,6 +23,11 @@
       state: directory
       mode: '0755'
 
+  - name: Override osp_interface for assisted installer deployments
+    set_fact:
+      osp_interface: enp8s0
+    when: ocp_ai|bool
+
   - name: Render templates to yaml dir
     template:
       src: "osp/networks/{{ item }}.j2"
--- a/ansible/ocp_ai_post_install.yaml
+++ b/ansible/ocp_ai_post_install.yaml
@@ -82,7 +82,7 @@
 
     - name: Wait for Metal3 pods to restart
       shell: |
-        oc wait pod --for condition=ready -n openshift-machine-api -l k8s-app=metal3 --timeout={{ default_timeout }}s
+        oc wait pod --for condition=ready -n openshift-machine-api -l k8s-app=metal3 --timeout={{ default_timeout*2 }}s
       environment:
         <<: *oc_env
 
--- a/ansible/ocp_ai_post_install.yaml
+++ b/ansible/ocp_ai_post_install.yaml
@@ -1,0 +1,86 @@
+---
+- hosts: localhost
+  vars_files: vars/default.yaml
+  roles:
+  - oc_local
+
+  tasks:
+  - name: Extra post-install steps for assisted installer
+    block:
+    - name: Include AI variables
+      include_vars: vars/ocp_ai.yaml
+
+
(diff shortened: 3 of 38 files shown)


ansible/install_networks.yaml
```diff
@@ -23,6 +23,11 @@
       state: directory
       mode: '0755'
 
+  - name: Override osp_interface for assisted installer deployments
+    set_fact:
+      osp_interface: enp8s0
```
dprince, line 28: should this be a variable, like it might change if we update the VMs right?

abays: Yes, that's doable.  I mean, you can just use `osp_interface` already, but it seems like we usually leave that as the default (`enp6s0`).  Was just trying to avoid the user having to figure out what it should be and to remember to override it manually in thir `local-defaults.yaml`.  Maybe we could just have another variable all-together (i.e. `ai_osp_interface`), but it would require more conditional checks in the playbooks then this simple `set_fact` override.

ansible/hosts
```diff
@@ -1,3 +1,6 @@
+[all:vars]
+ocp_ai=False
+
 localhost ansible_connection=local
```
stuggi, line 4: this needs to be outside the [all:vars] section, otherwise it will not use the local connection for localhost

stuggi (commented):
some comments to align vm configuration with what dev-scripts use, so we have the same disk naming.

ansible/ocp_ai_prep.yaml
```diff
@@ -0,0 +1,552 @@
...
+            sudo virt-install \
+            --virt-type=kvm \
+            --name "{{ ocp_cluster_name }}-master-$((i-1))" \
+            --memory {{ ocp_master_memory }} \
+            --vcpus={{ ocp_master_vcpu }} \
+            --os-variant=rhel8.2 \
+            --os-type linux \
+            --network=bridge:{{ ocp_cluster_name }}bm,mac="{{ ocp_ai_bm_bridge_master_mac_prefix }}$((i-1))" \
+            --network=bridge:{{ ocp_cluster_name }}pr,mac="{{ ocp_ai_prov_bridge_master_mac_prefix }}$((i-1))" \
+            --disk path=/var/lib/libvirt/images/{{ ocp_cluster_name }}-master-$((i-1)).qcow2,size={{ ocp_master_disk }},bus=virtio,format=qcow2 \
```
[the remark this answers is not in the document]
stuggi: this requires the ironic hints to be changed, right now we have {'name': 's== /dev/vda'

ansible/ocp_ai_prep.yaml
```diff
@@ -0,0 +1,552 @@
...
+        shell: |
+          virsh pool-define /tmp/storage-pool.xml
+          virsh pool-autostart default
+          virsh pool-start default
+      when: libvirt_default_pool.stdout == ""
+
+    - name: Create AI OCP VM XML definitions
+      shell: |
+        for i in {1..{{ ocp_num_masters }}}; do
+            sudo virt-install \
```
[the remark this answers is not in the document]
abays: This `sudo` and the others you mentioned can be removed.  I copied these scripts from my early non-Ansible approach, where I happened to need the `sudo`.  Since this playbook already runs as `root`, these are pointless anyhow.

ansible/ocp_ai_post_install.yaml
```diff
@@ -0,0 +1,89 @@
...
+      environment:
+        <<: *oc_env
+
+    - name: Pause for 15 seconds to allow Metal3 to begin reconciliation
+      pause:
+        seconds: 15
+
+    - name: Wait for Metal3 pods to restart
+      shell: |
+        oc wait pod --for condition=ready -n openshift-machine-api -l k8s-app=metal3 --timeout={{ default_timeout }}s
```
stuggi, line 85: Note, I hit once where the default_timeout = 240 was not enough and that task failed.

abays: Ack.  Will boost it

stuggi (approved):
/lgtm

PR #102, opened 2021-04-29
Use retries in wait tasks

stuggi:
Sometimes wait condition fail like:

Error from server: etcdserver: request timed out

Lets use retries to rerun the wait.


abays (approved):
/lgtm

PR #103, opened 2021-04-29
support multiple external dns server

stuggi:
Also increase the get_url timeout since it can take longer than
10s for the dns resolution to switch to the next server when the
targeted one is down.

### DIFF
--- a/ansible/vars/ocp_ai.yaml
+++ b/ansible/vars/ocp_ai.yaml
@@ -1,5 +1,9 @@
 ---
-ocp_ai_external_dns: 10.11.5.19
+ocp_ai_external_dns:
+  - 10.11.5.19
+  - 10.38.5.26
+  - 10.10.160.2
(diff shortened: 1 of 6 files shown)


ansible/vars/ocp_ai.yaml
```diff
@@ -1,5 +1,9 @@
 ---
-ocp_ai_external_dns: 10.11.5.19
+ocp_ai_external_dns:
+  - 10.11.5.19
+  - 10.38.5.26
+  - 10.10.160.2
+  - 10.5.30.160
```
stuggi, line 6: maybe we want to remove those here and replace them with public ones and overwrite in a privat vars file?

abays: Seems reasonable if we think enough people use this outside of Red Hat

abays (approved):
/lgtm

PR #104, opened 2021-04-29
Adjust Ansible README for AI

abays:


ansible/README.md
```diff
@@ -51,26 +74,16 @@ If using CI_TOKEN as shell variable, which has precedence over
...
+##### Assisted Installer (AI) install via assisted service
 
-`vars/rhel-subscription.yaml` in the format:
+Create `local-defaults.yaml` file with settings like so:
 
 ```
-rhel_subscription_activation_key: xyz
-rhel_subscription_org_id: "123123123"
+ocp_ai: true
+ocp_version: 4.7
```
stuggi, line 83: this is the default in `var/defaults.yaml`, but it is probably good to make sure it is that version

ansible/README.md
```diff
@@ -114,52 +127,68 @@ sshuttle -r <user>@<virthost> 192.168.111.0/24 192.168.25.0/24
...
+
 ```
 oc exec -it openstackclient -- cat /home/cloud-admin/tripleo-deploy/tripleo-overcloud-passwords.yaml | grep -w AdminPassword
 ```
 
-#### Cleanup full env:
+### Cleanup full env:
 
 ```
 make cleanup
```
stuggi, line 174: we don't have the `cleanup` target, maybe remove this for now. or we change it `openstack_cleanup` which deletes all osbms, osctlplane and ocnets

ansible/README.md
abays: Yeah, just for the time being until the Metal3 provisioning injection works for 4.6 and 4.8+ (maybe it does already).  The 4.7 "requirement" is just because that's the only version I was able to get PXE provisioning to work with AI

ansible/README.md
abays: Ack, will drop this

stuggi (approved):
/lgtm

PR #109, opened 2021-04-30
Use biosdev nic names for osvmset and osbmset 

stuggi:
To prevent nic re-ordering nic biosdev names should be used for
overcloud nodes.

Depends-On: https://github.com/openstack-k8s-operators/osp-director-operator/pull/217

### DIFF
--- a/ansible/datavolume.yaml
+++ b/ansible/datavolume.yaml
@@ -27,13 +27,20 @@
 
   - name: Get RHEL guest base image
     when: not stat_result.stat.exists
-    get_url:
-      url: "{{ osp_controller_base_image_url }}"
-      dest: "{{ osp_controller_base_image_url_path }}"
-      owner: ocp
-      group: ocp
-      mode: '0644'
-      timeout: 30
+    block:
+    - name: Download RHEL guest image {{ osp_controller_base_image_url }}
+      get_url:
+        url: "{{ osp_controller_base_image_url }}"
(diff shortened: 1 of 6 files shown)


ansible/install_computes.yaml
```diff
@@ -67,6 +68,31 @@
...
+  - name: Set path to RHEL base image for assisted installer
+    set_fact:
+      osp_vm_base_image_url_path: "{{ ocp_ai_http_store_dir | default('/opt/http_store/data', true) }}"
+    when: ocp_ai|bool
+
+  - name: Set fact for OSP compute image
+    set_fact:
+      osp_vm_image: "{{ osp_controller_base_image_url | basename }}"
+  - set_fact:
+      osp_bms_image: "compute-{{ osp_vm_image }}"
```
abays, line 85: This might fail linting.  We could just combine the facts into one task probably.

stuggi: We can not reference facts specified and used in the same task. I need to name the 2nd task

abays: Ah, understood

stuggi:
maybe we want to have biosdev names also used on the controller vms?

abays:
+1 for consistency across VMs/BMs if it's not too much trouble

stuggi:
> +1 for consistency across VMs/BMs if it's not too much trouble

done

PR #112, opened 2021-05-05
Align OSP intf names for IPI and AI

abays:
Changes AI VMs so that they have `enp7s0` for the OSP network interface, just like dev-scripts VMs now do


stuggi (approved):
/lgtm

PR #114, opened 2021-05-07
Move httpd port to 8081

abays:
Moves default `httpd` port for AI to 8081 instead of 8080 to avoid conflicts when running operator locally


stuggi (approved):
/lgtm

PR #117, opened 2021-05-12
Git support for the osplaybookgenerator, etc

dprince:
depends on https://github.com/openstack-k8s-operators/osp-director-operator/pull/241

### DIFF
--- a/ansible/git.yaml
+++ b/ansible/git.yaml
@@ -57,6 +57,7 @@
       PATH: /usr/local/bin:{{ ansible_env.PATH }}
 
   - name: git init playbooks.git
+    # noqa 303
     shell: |
        git init .
     args:
--- a/ansible/git.yaml
+++ b/ansible/git.yaml
@@ -68,15 +68,21 @@
 
 - hosts: convergence_base
   vars_files: vars/default.yaml
-  become: true
+  roles:
+  - oc_local
 
   tasks:
-    - name: copy git .ssh key to local_working_dir
+
+    - name: register git .ssh key
+      become: true
+      become_user: git
+      command: "cat /home/git/.ssh/id_rsa"
+      register: git_ssh_key
+
+    - name: copy git .ssh key to working_dir
       copy:
-        src: /home/git/.ssh/id_rsa
-        dest: "{{ local_working_dir }}/git_id_rsa"
-        owner: root
-        group: root
+        content: "{{ git_ssh_key.stdout }}"
+        dest: "{{ working_dir }}/git_id_rsa"
         mode: 0644
 
 - hosts: convergence_base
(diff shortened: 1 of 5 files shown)


ansible/openstack_cleanup.yaml
```diff
@@ -18,6 +18,11 @@
       - "oc delete -n openstack openstackplaybookgenerator --all"
       - "oc delete -n openstack openstackephemeralheat --all"
 
+  - name: create playbooks git repo dir
```
abays, line 21: `s/create/delete` ?

ansible/git.yaml
```diff
@@ -0,0 +1,88 @@
...
+    shell: >
+      oc get secret -n openstack git-secret --ignore-not-found
+    environment: &oc_env
+      PATH: "{{ oc_env_path }}"
+      KUBECONFIG: "{{ kubeconfig }}"
+    register: git_secret_exist_switch
+
+  - name: create secret for git ssh key
+    shell: |
+            oc create secret generic git-secret -n openstack --from-file=git_ssh_identity=/home/git/.ssh/id_rsa --from-literal=git_url=git@{{ ansible_default_ipv4.address }}:/home/git/playbooks.git
```
abays, line 85: Do we typically use

```
[convergence_base]
localhost
```

...in the hosts file for our playbooks?  Just wondering what `ansible_default_ipv4.address` holds, i.e. if it would put `127.0.0.1` in the secret in such a case.

ansible/openstack_cleanup.yaml
dprince: Just pushed a fix for this. Thanks

ansible/git.yaml
dprince: Well, when I run it I get a 10. IP (the one that I want). So this functionally works. Perhaps @mpryc has some thoughts on this too though

mpryc, line 85: CI gave -1 to that:
 error: error reading /home/git/.ssh/id_rsa: permission denied

It's because tasks in the CI are run as different user.

mpryc: ansible_default_ipv4.address is the same as IP from:
`ip route show|grep default`

I think the best is to include as well ansible_connection, this ensures tasks are executed locally.

`[convergence_base]`
`localhost ansible_connection=local`

Here is some more information about it:
  https://docs.ansible.com/ansible/latest/inventory/implicit_localhost.html#implicit-localhost

dprince: The latest commit copies the SSH key to the local workdir first and then uses it from there. I think this should solve the CI issue.

mpryc:
/lgtm

Added dependency on the osp-director-operator #241 as it's not possible to install without getting #241 first in the quay.io. The CI fix worked, thanks.

`TASK [Start playbook generator] ************************************************
Thursday 20 May 2021  21:57:30 +0000 (0:00:00.661)       0:00:06.654 ********** 
fatal: [localhost]: FAILED! => {
    "changed": true,
    "cmd": "set -e\noc apply -n openstack -f \"/home/rhos-ci/ostest-working/yamls/playbook_generator/openstackplaybookgenerator.yaml\"\n",
    "delta": "0:00:00.929692",
    "end": "2021-05-20 21:57:31.818687",
    "rc": 1,
    "start": "2021-05-20 21:57:30.888995"
}
STDERR:
The OpenStackPlaybookGenerator "default" is invalid: spec.openstackClientName: Required value`

PR #121, opened 2021-05-14
Fix AI install service

abays:
Basic fix to get AI working again.  Disabling fine-grained version selection for AI for the moment, just to make sure the stock defaults from upstream work first.


mpryc:
Depends on #120

mpryc:
/lgtm

We can ignore pr-merge result as it was broken when this PR was introduced.

PR #123, opened 2021-05-17
Register SSH public key before using in JINJA script

mpryc:
Use of lookup plugin inside JINJA template as different user fails
with error that file was not found. As a work-around we register
this file in previous task and pass it to JINJA as variable instead.

This allows to get SSH public key as any user that is specified in the
Ansible task.


abays (approved):
/lgtm

PR #124, opened 2021-05-17
Universal destroy_ocp

abays:
CI improvement that allows `make destroy_ocp` to agnostically clean whatever previous OCP cluster might exist (regardless of whether IPI or AI)

### DIFF
--- a/ansible/ocp_ai_destroy.yaml
+++ b/ansible/ocp_ai_destroy.yaml
@@ -84,6 +84,7 @@
       ./stop-assisted-service.sh
     args:
       chdir: "{{ base_path }}/assisted-service-onprem"
+    ignore_errors: true
 
 
   ### BRIDGES
--- a/ansible/ocp_installer_destroy.yaml
+++ b/ansible/ocp_installer_destroy.yaml
@@ -1,11 +1,9 @@
 ---
 - name: Destroy the dev-scripts OCP cluster
   import_playbook: ocp_dev_scripts_destroy.yaml
-  when: not (ocp_ai|bool)
 
 - name: Destroy the assisted installer OCP cluster
   import_playbook: ocp_ai_destroy.yaml
-  when: ocp_ai|bool
 
 - name: Cleanup libvirt resources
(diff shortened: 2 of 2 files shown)


dprince:
universal destruction! whats not to like here...

One question though. Doesn't this put us at risk of breakage if for example dev-scripts cleanup is broken again (it was just recently broken for example).

abays:
> universal destruction! whats not to like here...
> 
> One question though. Doesn't this put us at risk of breakage if for example dev-scripts cleanup is broken again (it was just recently broken for example).

That's actually a good point.  Perhaps we should add an `ignore_errors: true` to the `import_playbook` for `'ocp_dev_scripts_destroy.yaml`.  Thoughts?

mpryc:
> > universal destruction! whats not to like here...
> > One question though. Doesn't this put us at risk of breakage if for example dev-scripts cleanup is broken again (it was just recently broken for example).
> 
> That's actually a good point. Perhaps we should add an `ignore_errors: true` to the `import_playbook` for `'ocp_dev_scripts_destroy.yaml`. Thoughts?

From my perspective if something is broken I would prefer it to fail rather then silently hide problem. If we ignore broken cleanup this probably will affect next job in the CI queue and the job will fail anyway resulting us to spend more time on finding why the job fail then having direct information that environment was not properly cleaned up before starting new job.

abays:
> > > universal destruction! whats not to like here...
> > > One question though. Doesn't this put us at risk of breakage if for example dev-scripts cleanup is broken again (it was just recently broken for example).
> > 
> > 
> > That's actually a good point. Perhaps we should add an `ignore_errors: true` to the `import_playbook` for `'ocp_dev_scripts_destroy.yaml`. Thoughts?
> 
> From my perspective if something is broken I would prefer it to fail rather then silently hide problem. If we ignore broken cleanup this probably will affect next job in the CI queue and the job will fail anyway resulting us to spend more time on finding why the job fail then having direct information that environment was not properly cleaned up before starting new job.

Okay, let's keep it as it is for now

PR #125, opened 2021-05-17
Allow OCP user to write to AI http_store

abays:
Now that we download the RHEL image as `ocp` user, we need to open up the perms on the AI httpd image store


stuggi (approved):
/lgtm

PR #127, opened 2021-05-18
Add default AI HW validation JSON

abays:
Another upstream AI change was recently made that requires a `HW_VALIDATOR_REQUIREMENTS` env var to be set in the AI service container.  For now we'll inject until they respin the associated `stable`-tag AI image that includes it by default.


mpryc:
CI passed the failing part.

/lgtm

PR #128, opened 2021-05-18
Local yaml directories should be owned by ansible user

mpryc:
Directories to store CR yaml files should be owned by the user
who is invoking ansible playbook as the resulting dir is inside
that user home directory.


mpryc:
This passed CI and also allows to use AI in the CI. Ready to review & merge.

abays (approved):
/lgtm

PR #131, opened 2021-05-24
Allow AI to work for RHEL 8.4

abays:
Something is different in RHEL 8.4 vs 8.3 (and earlier), such that `sushy-tools` configuration needs to point to a different bootloader for `x86_64` UEFI

### DIFF
--- a/ansible/templates/ai/sushy-tools/sushy-emulator.conf.j2
+++ b/ansible/templates/ai/sushy-tools/sushy-emulator.conf.j2
@@ -24,7 +24,11 @@ SUSHY_EMULATOR_LIBVIRT_URI = u'qemu:///system'
 # system architecture
 SUSHY_EMULATOR_BOOT_LOADER_MAP = {
     u'UEFI': {
+{% if ansible_distribution_version|float <= 8.3 %}
         u'x86_64': u'/usr/share/edk2/ovmf/OVMF_CODE.secboot.fd',
+{% else %}
+        u'x86_64': u'/usr/share/OVMF/OVMF_CODE.secboot.fd',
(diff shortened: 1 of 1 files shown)


ansible/templates/ai/sushy-tools/sushy-emulator.conf.j2
```diff
@@ -24,7 +24,11 @@ SUSHY_EMULATOR_LIBVIRT_URI = u'qemu:///system'
 # system architecture
 SUSHY_EMULATOR_BOOT_LOADER_MAP = {
     u'UEFI': {
+{% if ansible_distribution_version|float <= 8.3 %}
         u'x86_64': u'/usr/share/edk2/ovmf/OVMF_CODE.secboot.fd',
```
mpryc, line 28: On rhel8.3 it's a link so maybe keeping just  /usr/share/OVMF/OVMF_CODE.secboot.fd will work?

# ls -lah /usr/share/OVMF/OVMF_CODE.secboot.fd 
lrwxrwxrwx. 1 root root 33 Nov 23  2020 /usr/share/OVMF/OVMF_CODE.secboot.fd -> ../edk2/ovmf/OVMF_CODE.secboot.fd

abays: I no longer have a RHEL 8.3 machine to test that, but if it works, let's do that instead of the if-conditional.

mpryc: Based on the link/file expectation I think your fix is fine.
/lgtm

PR #134, opened 2021-05-27
Make sure git_ssh_key has EOL

stuggi:
the copy ansible module will not add EOL to the last line when
adding git_ssh_key.stdout as the content to the git_ssh_key file.
This makes sure we add a EOL to at the end

### DIFF
--- a/ansible/git.yaml
+++ b/ansible/git.yaml
@@ -80,8 +80,9 @@
       register: git_ssh_key
 
     - name: copy git .ssh key to working_dir
+      #command: "echo {{ git_ssh_key.stdout }} > {{ working_dir }}/git_id_rsa"
       copy:
-        content: "{{ git_ssh_key.stdout }}"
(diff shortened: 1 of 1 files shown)


ansible/git.yaml
```diff
@@ -80,8 +80,9 @@
       register: git_ssh_key
 
     - name: copy git .ssh key to working_dir
+      #command: "echo {{ git_ssh_key.stdout }} > {{ working_dir }}/git_id_rsa"
       copy:
-        content: "{{ git_ssh_key.stdout }}"
+        content: "{{ git_ssh_key.stdout }}{{ '\n' }}"
```
stuggi, line 85: just adding \n won't work as it is interpreted as a string, the workaround I found is to add {{ '\n' }} to the end which is  jinja2

mpryc: Another option would be to change line 79 to:
cat /home/git/.ssh/id_rsa && echo

PR #136, opened 2021-06-01
Update to use current latest 16.2 tripleoclient image

stuggi:
The latest image now contain the ceph-ansible packages. For now
added the image to our quay.io repo to not have to add internal
registry ssl certs.

Depends-On: https://github.com/openstack-k8s-operators/osp-director-operator/pull/255


abays (approved):
/lgtm

PR #137, opened 2021-06-02
Use single shell block for Hack to fix dev-scripts

mpryc:
Using shell to attach pci devices for dev-scripts in a single
block rather then split between two tasks.

### DIFF
--- a/ansible/ocp_vm_setup_extra_nics.yaml
+++ b/ansible/ocp_vm_setup_extra_nics.yaml
@@ -81,21 +81,19 @@
       with_items: "{{ worker_ocp_inactive.stdout_lines[0].split(' ') }}"
       ignore_errors: true
 
-    - name: HACK to fix dev-scripts bug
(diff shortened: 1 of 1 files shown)


abays (commented):
This looks good, but could we change the `name` of the two blocks so that one says something like `Attach the osp network to ACTIVE worker VM's for dev-scripts` and the other mentions AI?

mpryc:
> This looks good, but could we change the `name` of the two blocks so that one says something like `Attach the osp network to ACTIVE worker VM's for dev-scripts` and the other mentions AI?

Changed, however I think it's better to not mention the AI as it's obvious from the task run (skipped as conditions are not met). The other task is "Attach the osp network to INACTIVE worker VM's" so it also does not mention if it's AI or dev-tools. Of course if you want AI in the name I will add it :)

abays (approved):
/lgtm

PR #142, opened 2021-06-10
Remove Ceph OSD Journal Size

fultonj:
The journal_size parameter applies to Filestore but the template
deploys with Bluestore so remove the unnecessary paramter.

Signed-off-by: John Fulton <fulton@redhat.com>


stuggi (approved):
/lgtm

PR #145, opened 2021-06-15
use rhel-guest-image 8.4 GA image version

stuggi:
until now we used 8.4 beta image. Lets use a GA image version.


dprince (approved):

dprince:
all good from me once CI passes

PR #149, opened 2021-06-18
Retry for the datavolume wait task

mpryc:
The oc wait datavolume fails randomly, however after manual
check the service is always available. Adding retry, so
we can limit number of unexpected deploy errors.

### DIFF
--- a/ansible/datavolume.yaml
+++ b/ansible/datavolume.yaml
@@ -55,13 +55,25 @@
     register: datavolume_switch
 
   - name: Deploy the datavolume
-    shell: |
-      virtctl image-upload dv openstack-base-img -n {{ namespace }} --size={{ osp_controller_disk_size }}G --image-path={{ osp_controller_base_image_url_path }} --insecure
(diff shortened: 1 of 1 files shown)


abays (approved):
/lgtm

ansible/datavolume.yaml
```diff
@@ -55,13 +55,25 @@
...
     environment:
       <<: *oc_env
-    when: datavolume_switch.rc == 1
+    block:
+      - name: Upload OSP Controller base image
+        shell: |
+          virtctl image-upload dv openstack-base-img -n {{ namespace }} --size={{ osp_controller_disk_size }}G --image-path={{ osp_controller_base_image_url_path }} --insecure
+    rescue:
+      - name: Remove pod from failed OSP Controller upload
+        shell: oc delete pod cdi-upload-openstack-base-img -n {{ namespace }}
```
dprince, line 67: shouldn't this just delete the resulting 'DataVolume' (dv) instead?

abays (approved):
/lgtm

PR #152, opened 2021-06-24
Retry Install OpenShift Virtualization Operator if fails

mpryc:
It happens that the task fails randomly with the error
"no endpoints available for service "hco-webhook-service"

Retry does fix the above issue most of the times.


stuggi (approved):
/lgtm

PR #157, opened 2021-06-30
Increase time to allow osplaybookgenerator to finish

mpryc:
Seems like 720s is not enough to allow osplaybookgenerator to
finish.


stuggi (approved):
/lgtm

PR #158, opened 2021-07-01
bump osp tags to 16.2_20210630.1

stuggi:


abays:
Changes seem fine to me, but CI is failing because the compute is unreachable via SSH on ctlplane network.  Is this just a hiccup?

PR #161, opened 2021-07-01
Make RedHat subscription tasks robust

mpryc:
The RedHat subscription manager is failing randomly either on
the unsubscribe or subscribe tasks. Adding retries make it more
robust. Unsubscribe failing task can be ignored with the follow
change to allow registering system even if it was already
registered.


stuggi (approved):
/lgtm

PR #162, opened 2021-07-01
Extend tempest tests for current phase2 tests

stuggi:
Adds tempest vars to control which tests to run.
tempest_smoketest run tempest with --smoke, default
to false.
If tempest_smoketest is false, information from tempest_test_dict
is used to run tempest with --regex, --include-list, and
--exclude-list

current test list from https://github.com/redhat-openstack/infrared/blob/master/plugins/tempest/vars/tests/full.yml

Depends-On: https://github.com/openstack-k8s-operators/osp-director-operator/pull/311

### DIFF
--- a/ansible/Makefile
+++ b/ansible/Makefile
@@ -47,6 +47,11 @@ post_install: hosts local-defaults.yaml
 	ocp_image_registry.yaml \
 	ocp_vm_setup_extra_nics.yaml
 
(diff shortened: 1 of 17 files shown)


abays (approved):
/lgtm

stuggi:
Added commit which introduce separate external network + nic for VMs

abays (approved):
/lgtm

PR #167, opened 2021-07-08
Drop director_operator_version (use csv_version instead)

dprince:
As far as I know we always set the tag name to match the CSV version
in our development environments.


stuggi (approved):
/lgtm

PR #172, opened 2021-07-13
AI and IPI use same image server URLs

abays:
Consolidates AI and IPI deployments to use the same URL structure for RHEL image locations.  In doing so, AI no longer supports customizable provisioning and baremetal network CIDRs, and instead uses the same forced defaults of the IPI dev-scripts approach.

### DIFF
--- a/ansible/templates/osp/compute/openstackbaremetalset.yaml.j2
+++ b/ansible/templates/osp/compute/openstackbaremetalset.yaml.j2
@@ -11,11 +11,7 @@ spec:
   # How many nodes to provision
   count: {{ osp_compute_count }}
   # The image to install on the provisioned nodes
-{% if not (ocp_ai|bool) %}
-  baseImageUrl: http://172.22.0.1/images/{{ osp_controller_base_image_url | basename }}
-{% else %}
-  baseImageUrl: http://{{ ocp_ai_bm_ip }}:{{ ocp_ai_http_store_port | default('8081', true) }}/{{ osp_controller_base_image_url | basename }}
-{% endif %}
+  baseImageUrl: http://192.168.111.1/images/{{ osp_controller_base_image_url | basename }}
   provisionServerName: openstack
   # The secret containing the SSH pub key to place on the provisioned nodes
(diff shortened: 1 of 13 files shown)


ansible/templates/osp/compute/openstackbaremetalset.yaml.j2
```diff
@@ -11,11 +11,7 @@ spec:
   # How many nodes to provision
   count: {{ osp_compute_count }}
   # The image to install on the provisioned nodes
-{% if not (ocp_ai|bool) %}
-  baseImageUrl: http://172.22.0.1/images/{{ osp_controller_base_image_url | basename }}
-{% else %}
-  baseImageUrl: http://{{ ocp_ai_bm_ip }}:{{ ocp_ai_http_store_port | default('8081', true) }}/{{ osp_controller_base_image_url | basename }}
-{% endif %}
+  baseImageUrl: http://192.168.111.1/images/{{ osp_controller_base_image_url | basename }}
```
abays, line 14: This works for both AI and IPI, because in both cases the Apache container binds on port 80 on the provisioning host on all interfaces.

ansible/templates/osp/compute/openstackprovisionserver.yaml.j2
```diff
@@ -5,8 +5,4 @@ metadata:
   namespace: openstack
 spec:
   port: 8080
-{% if not (ocp_ai|bool) %}
-  baseImageUrl: http://172.22.0.1/images/{{ osp_controller_base_image_url | basename }}
-{% else %}
-  baseImageUrl: http://{{ ocp_ai_bm_ip }}:{{ ocp_ai_http_store_port | default('8081', true) }}/{{ osp_controller_base_image_url | basename }}
-{% endif %}
+  baseImageUrl: http://192.168.111.1/images/{{ osp_controller_base_image_url | basename }}
```
abays, line 8: This works for both AI and IPI, because in both cases the Apache container binds on port 80 on the provisioning host on all interfaces.

stuggi (approved):
/lgtm

PR #174, opened 2021-07-13
Enable additional tempest features

stuggi:
This allows to enable/disable tempest features using
tempest_enable_feature_dict and tempest_disable_feature_dict.

It also:
* enables the following compute-feature-enabled features
in our default vars file: vnc_console, live_migration,
block_migration_for_live_migration, volume_backed_live_migration
and console_output
* deployes 2 OSP compute nodes per default that live migration
tests can be done.


abays (approved):
/lgtm

PR #180, opened 2021-07-14
Add task to configure custom OCP registries

dprince:


ansible/ocp_custom_registries.yaml
```diff
@@ -0,0 +1,33 @@
...
+    shell: |
+      set -e
+
+      oc patch image.config.openshift.io/cluster --type merge \
+        --patch '{"spec":{"registrySources":{"allowedRegistries":["registry.redhat.io","quay.io","registry-proxy.engineering.redhat.com","gcr.io"],"insecureRegistries":["registry-proxy.engineering.redhat.com"]}}}'
+    environment:
+      PATH: "{{ oc_env_path }}"
+      KUBECONFIG: "{{ kubeconfig }}"
+
+  - name: wait for the machine config pools to update for image registry customizations
```
dprince, line 21: Tried this several times and it seems to take almost 15 minutes to complete. During that time it seems to cause intermittent disruptions to the API (thus the retry). Perhaps we don't use this by default, but those who want to test with downstream/release packages can just run the extra task. If we have a CI job for the release packages we could then just call this extra task after OCP is installed.

PR #186, opened 2021-07-19
Support for OCP 4.8.2 with AI

dprince:


abays:
I got this to work all the way to full overcloud deployment, but it required a couple extra changes.  First, I needed to change this line:

https://github.com/openstack-k8s-operators/osp-director-dev-tools/blob/49994344da3702b7393da23d4109cdb73c641ece/ansible/vars/ocp_ai.yaml#L2

...and set it to `ocp_ai_discovery_iso: https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/pre-release/4.8.0-rc.3/rhcos-live.x86_64.iso` (or override in `local-defaults.yaml`).  

Then I needed to hack this line:

https://github.com/openstack-k8s-operators/osp-director-dev-tools/blob/49994344da3702b7393da23d4109cdb73c641ece/ansible/ocp_ai.yaml#L89

...because `latest-4.8` OCP client does not exist yet.  Instead I changed the line so it produced `candidate-4.8`.  This could, of course, be done more elegantly.

ansible/host_prep.yaml
```diff
@@ -96,9 +96,9 @@
           msg: RHEL 8.3+ is required for the assisted installer deployments
         when: ansible_distribution != "RedHat" or ansible_distribution_version|float < 8.3
 
-      - fail:
-          msg: "OCP version {{ ocp_version }} is not allowed.  Only OCP 4.7 is currently supported for assisted installer deployments."
-        when: ocp_version != 4.7
+      #- fail:
+          #msg: "OCP version {{ ocp_version }} is not allowed.  Only OCP 4.7 is currently supported for assisted installer deployments."
+        #when: ocp_version != 4.7
```
abays, line 101: Maybe we could change this condition to `ocp_version >= 4.7`

ansible/ocp_ai.yaml
```diff
@@ -86,7 +86,9 @@
   tasks:
   - name: Get the ocp client tar gunzip file
     get_url:
-      url: "https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest-{{ ocp_version }}/openshift-client-linux.tar.gz"
+      #url: "https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest-{{ ocp_version }}/openshift-client-linux.tar.gz"
+      # FIXME: switch to latest- when available
+      url: "https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.8.2/openshift-client-linux.tar.gz"
```
dprince, line 91: I *think* this will get updated to latest-4.8 once it goes GA next week. If it does then I can change it back to the old format.

ansible/host_prep.yaml
dprince: @abays wouldn't we want it to fail only if the version is <= 4.6

abays: Sorry, yes.  I was thinking backwards. :-)

dprince: got it, just push a commit to reactivate this validation

ansible/ocp_ai.yaml
abays: If `ocp_version` is set to 4.7, it will use this same binary, right?  Should we have some sort of conditional logic here to get the 4.7 or 4.8 versions depending on what was chosen?  On the other hand, it's likely that both `oc` binaries will work with either version (that has been my experience).

dprince: Gotcha, Just pushed a patch to conditionally download the client tools based on the URL (safer). But yeah I haven't seen an issue in mix/matching client version binaries for what we are doing so far, but something may come up.

abays:
Looks good as-is to me.  +1 to merge or we can wait if you have more updates incoming.

abays (approved):

PR #189, opened 2021-07-29
Allow to install goloang when executed as non-root

mpryc:
Allow to install yum packages when executed as non-root user.


stuggi (approved):
/lgtm

abays (approved):
/lgtm

PR #192, opened 2021-07-30
Bump to CNV version 4.8

dprince:


abays (approved):
/lgtm

stuggi:
since we have not set to use OCP 4.8 or later per default, this is probably no longer needed? maybe we add var files per OCP release, like keep 4.7 in default, but have a 4.9 var file to overwrite the OCP version settings?

PR #193, opened 2021-08-02
Add retry for image reg availability

abays:
Sometimes the check for image registry availability fails temporarily right after the deployment has finished.  Let's add a retry to give the `wait` condition a few chances to succeed.

### DIFF
--- a/ansible/ocp_image_registry.yaml
+++ b/ansible/ocp_image_registry.yaml
@@ -77,3 +77,7 @@
     environment:
       PATH: "{{ oc_env_path }}"
       KUBECONFIG: "{{ kubeconfig }}"
+    register: image_registry_available
+    until: image_registry_available is not failed
+    retries: "{{ (default_timeout / 5)|int }}"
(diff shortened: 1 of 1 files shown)


ansible/ocp_image_registry.yaml
```diff
@@ -77,3 +77,7 @@
     environment:
       PATH: "{{ oc_env_path }}"
       KUBECONFIG: "{{ kubeconfig }}"
+    register: image_registry_available
+    until: image_registry_available is not failed
+    retries: "{{ (default_timeout / 5)|int }}"
```
mpryc, line 82: I don't think this is ideal. We do have "--timeout={{ default_timeout }}s" in the oc wait, then here the same timeout just divided by 5, so if the default timeout is 50s then
oc wait 50s * 10 times + 50s (from delay), this increases timeout quite a lot ?

mpryc:
as @abays explained similar logic is used elsewhere /lgtm

PR #195, opened 2021-08-05
Clean AI podman images

abays:
Removes AI container images during `make destroy_ocp` to ensure the next run always has the latest images


stuggi (approved):
/lgtm

PR #196, opened 2021-08-10
Remove AI Postgres data during cleanup

abays:
No reason to leave this data lingering after a clean-up

### DIFF
--- a/ansible/ocp_ai_destroy.yaml
+++ b/ansible/ocp_ai_destroy.yaml
@@ -86,6 +86,11 @@
       chdir: "{{ base_path }}/assisted-service-onprem"
     ignore_errors: true
 
+  - name: Remove Postgres data store
+    file:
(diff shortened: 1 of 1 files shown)


ansible/ocp_ai_destroy.yaml
```diff
@@ -86,6 +86,11 @@
       chdir: "{{ base_path }}/assisted-service-onprem"
     ignore_errors: true
 
+  - name: Remove Postgres data store
+    file:
+      path: "{{ ocp_ai_service_store_dir }}/data/postgresql"
```
mpryc, line 91: CI: The task includes an option with an undefined variable. The error was: 'ocp_ai_service_store_dir' is undefined

abays: Ah, thanks.  I forgot to default it.

PR #197, opened 2021-08-11
Initial osp17 enablement

stuggi:


stuggi:
from continuous-integration tests, deployment was successful. Tempest timed out/or had failures on subset of tests. merging it to allign with osp17 patch we landed on the operator side

PR #202, opened 2021-08-18
Preliminary Kubevirt fencing support 

abays:
This logic injects fencing bits into the overcloud image and controllers so that we can test it in our dev envs.  The real solution involves changes to the operator and also to upstream OpenStack bits.


ansible/osp_tripleo_puppet_overrides.yaml
```diff
@@ -0,0 +1,72 @@
...
+      oc get osnets -n openstack ctlplane -o jsonpath='{range .status.roleReservations.Controller.reservations[*]}{.ip}{"\n"}{end}'
+    environment: &oc_env
+      PATH: "{{ oc_env_path }}"
+      KUBECONFIG: "{{ kubeconfig }}"
+    register: ctrl_vm_ips
+
+  - name: Pre-install certain Puppet modules on controller VMs
+    shell: |
+      #!/bin/bash
+      oc rsh -n openstack openstackclient ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no cloud-admin@{{ item }} sudo dnf install -y puppet-tripleo puppet-pacemaker
```
stuggi, line 20: As we have the rendered ansible playbooks, we also have the inventory of the overcloud in the tripleoclient.
Therefore we could just use ansible to install, or run tasks on the controller nodes.

```
sh-4.4$ ansible -i /home/cloud-admin/playbooks/tripleo-ansible/tripleo-ansible-inventory.yaml -m ping Controller
controller-0 | SUCCESS => {
    "ansible_facts": {
        "discovered_interpreter_python": "/usr/libexec/platform-python"
    },
    "changed": false,
    "ping": "pong"
}
```
or copy a small playbook to the openstackclient pod to run all tasks on the Controller group at once, like we do to register the nodes [1]
[1] https://github.com/openstack-k8s-operators/osp-director-dev-tools/blob/master/ansible/templates/osp/rhsm.yaml.j2

ansible/osp_tripleo_puppet_overrides.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+  - name: Copy kubeconfig to openstackclient
+    shell: |
+      #!/bin/bash
+      oc cp -n openstack {{ kubeconfig }} openstackclient:/home/cloud-admin/kubeconfig
+      oc rsh -n openstack openstackclient chmod 0777 /home/cloud-admin/kubeconfig
+    environment:
+      <<: *oc_env
+
+  - name: Copy Puppet fencing files and kubeconfig from openstackclient to controller VMs
```
stuggi, line 62: those tasks could also run via ansible as mentioned above

ansible/osp_tripleo_puppet_overrides.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+  tasks:
+  - name: Get VM control plane IPs
+    shell: |
+      #!/bin/bash
+      oc get osnets -n openstack ctlplane -o jsonpath='{range .status.roleReservations.Controller.reservations[*]}{.ip}{"\n"}{end}'
+    environment: &oc_env
+      PATH: "{{ oc_env_path }}"
+      KUBECONFIG: "{{ kubeconfig }}"
+    register: ctrl_vm_ips
```
stuggi, line 15: with using ansible as described bellow I think this is then not required.

ansible/files/fencing/fencing.pp
```diff
@@ -0,0 +1,176 @@
...
+    create_resources('pacemaker::stonith::fence_ipmilan', $ipmilan_devices, $common_params)
+
+    $kdump_devices = local_fence_devices('fence_kdump', $all_devices)
+    create_resources('pacemaker::stonith::fence_kdump', $kdump_devices, $common_params)
+
+    $rhev_devices = local_fence_devices('fence_rhevm', $all_devices)
+    create_resources('pacemaker::stonith::fence_rhevm', $rhev_devices, $common_params)
+
+    $kubevirt_devices = local_fence_devices('fence_kubevirt', $all_devices)
+    create_resources('pacemaker::stonith::fence_kubevirt', $kubevirt_devices, $common_params)
```
stuggi, line 141: this is the only change to the file provided by the rpm, right? we could also patch the file provided by puppet-pacemaker in bellow ansible task

abays: This is the only change I made to get this working.  I don't know if anything else might be required, to be honest.

ansible/osp_tripleo_puppet_overrides.yaml
abays: Ack

ansible/osp_tripleo_puppet_overrides.yaml
abays: Sure, let me try that out and update the PR

ansible/osp_tripleo_puppet_overrides.yaml
```diff
@@ -0,0 +1,72 @@
...
+    - "fencing.pp"
+    - "fence_kubevirt.pp"
+
+  - name: Copy kubeconfig to openstackclient
+    shell: |
+      #!/bin/bash
+      oc cp -n openstack {{ kubeconfig }} openstackclient:/home/cloud-admin/kubeconfig
+      oc rsh -n openstack openstackclient chmod 0777 /home/cloud-admin/kubeconfig
+    environment:
+      <<: *oc_env
```
stuggi, line 60: I think it would make sense to add deployment instructions/or automation in the operator to create a dedicated user/kubeconfig for fencing (only allow fencing of VMs in the openstack namespace) and inject it into the VM using a secret [1], or cloud-init in [2]]? The obvious advantage of using a secret is that if for some reason the kubeconfig changes/has to change it is also updated in the VM, while using cloud-init there is the need to update it manually.

[1] https://kubevirt.io/2018/new-volume-types.html
[2] https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/templates/vmset/cloudinit/userdata

ansible/datavolume.yaml
```diff
@@ -47,6 +47,51 @@
...
+              regexp: "^API_VERSION='kubevirt.io/v1'"
+              line: "API_VERSION='kubevirt.io/v1alpha3'"
+
+          - name: Copy Kubevirt fencing agent code into {{ osp_controller_base_image_url_path }}
+            shell: |
+              #!/bin/bash
+              set -e
+              virt-copy-in -a "{{ osp_controller_base_image_url_path }}" \
+                "{{ base_path }}/fence_kubevirt" /usr/sbin
+
```
stuggi, line 94: we could also add this in the later steps where we get the puppet files in?

stuggi: I mean as part of the puppet overwrites playbook

ansible/osp_tripleo_puppet_overrides.yaml
abays: Yes, we will need to do something like that for the real thing.  The current hack which adds the full admin kubeconfig to the controllers is a terrible security risk.  Ideally we want a kubeconfig that just allows a minimal set of permissions for `VirtualMachine` resources in the `openstack` namespace.

ansible/datavolume.yaml
```diff
@@ -47,6 +47,51 @@
...
+            lineinfile:
+              path: "{{ base_path }}/fence_kubevirt"
+              regexp: '^sys.path.append\("@FENCEAGENTSLIBDIR@"\)'
+              line: 'sys.path.append("/usr/share/fence")'
+
+          - name: Replace Kubevirt API version in Kubevirt fencing agent code
+            lineinfile:
+              path: "{{ base_path }}/fence_kubevirt"
+              regexp: "^API_VERSION='kubevirt.io/v1'"
+              line: "API_VERSION='kubevirt.io/v1alpha3'"
```
stuggi, line 86: instead of fixed API_VERSION, would be good if this is a parameter in the fencing agent?

abays: That's certainly a question worth posing to the authors of the agent, I think.

stuggi (approved):
/lgtm

PR #203, opened 2021-08-19
Add baremetal worker

akiselev1:
Add an option to deploy a baremetal worker node in addition to virtual workers.

### DIFF
--- a/ansible/ocp_ai_destroy.yaml
+++ b/ansible/ocp_ai_destroy.yaml
@@ -112,6 +112,7 @@
       with_items:
         - bm
         - pr
+    ignore_errors: true
 
   ### remove osp_controller_base_image
 
--- a/ansible/ocp_ai_prep.yaml
+++ b/ansible/ocp_ai_prep.yaml
@@ -96,6 +96,11 @@
           ip4: "192.168.111.1/24"
           state: present
 
+      - name: Add BM interface to BM bridge
+        shell: |
+          /usr/bin/nmcli con add type bridge-slave ifname {{ ocp_bm_int }} master {{ ocp_cluster_name }}bm
+        when: ocp_bm_int is defined
+
       - name: Create provisioning bridge
         community.general.nmcli:
           conn_name: "{{ ocp_cluster_name }}pr"
--- a/ansible/vars/default.yaml
+++ b/ansible/vars/default.yaml
@@ -47,6 +47,18 @@ ocp_worker_memory: 30000
 ocp_worker_vcpu: 8
 ocp_worker_disk: 50
 
+# reserved for future expansion. Can be 0 or 1 at this point
+ocp_num_bm_workers: 0
+
+# required if ocp_num_bm_workers == 1
+#ocp_bm_worker_vendor: Dell
+#ocp_bm_worker_bmc_address: 10.16.231.76
+#ocp_bm_worker_bmc_user: root
+#ocp_bm_worker_bmc_password: calvin
+#ocp_bm_worker_mac: 40:a6:b7:2b:19:01
+#ocp_bm_worker_ip: 192.168.111.30
+#ocp_bm_int: ens3f0
+
 # OCP cluster name
 ocp_cluster_name: ostest
 #ocp_domain_name: defaults to test.metalkube.org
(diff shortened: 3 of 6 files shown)


ansible/ocp_ai_destroy.yaml
```diff
@@ -112,6 +112,7 @@
       with_items:
         - bm
         - pr
+    ignore_errors: true
```
abays, line 115: Curious as to what error you are seeing here.  Is it because there is an interface attached to the bridge?

akiselev1: This may be a separate issue. What I see is:
```

TASK [Delete existing bridges (if any)] **********************************************************************************************************************************************************************
Friday 20 August 2021  16:39:37 +0000 (0:00:00.230)       0:00:36.549 ********* 
failed: [localhost] (item=bm) => {"ansible_loop_var": "item", "changed": false, "item": "bm", "msg": "Error: unknown connection 'ostestbm'.\nError: cannot delete unknown connection(s): 'ostestbm'.\n", "name": "No Connection named ostestbm exists", "rc": 10}
ok: [localhost] => (item=pr)

[root@r740xd-l3c14u15 ~]# nmcli d
DEVICE          TYPE      STATE                   CONNECTION  
ens3f1          ethernet  connected               ens3f1      
cni-podman0     bridge    connected               cni-podman0 
external        bridge    connected               external    
ospnetwork      bridge    connected               ospnetwork 
```

abays: Weird.  I wonder where the `ostestbm` bridge went?  Perhaps you had run this playbook before and it destroyed the bridge then but then failed at some other step, and then you ran it again?  Anyhow, shouldn't be a problem to have the ignore here.

ansible/ocp_ai_prep.yaml
```diff
@@ -96,6 +96,11 @@
           ip4: "192.168.111.1/24"
           state: present
 
+      - name: Add BM interface to BM bridge
+        shell: |
+          /usr/bin/nmcli con add type bridge-slave ifname {{ ocp_bm_int }} master {{ ocp_cluster_name }}bm
```
abays, line 101: If for some reason the user sets `ocp_num_bm_workers` to 1 but doesn't set the required vars you mentioned in `vars/defaults.yaml`, this will throw an error because `ocp_bm_int` will be undeclared.  That being said, there's really no proper default value for this variable (as the interface will be different, likely, between environments), so even declaring it with a default will just lead to another error.  Ultimately it is user error to fail to follow instructions, so I think this is fine as-is.  Just wanted to point that out.

ansible/vars/default.yaml
```diff
@@ -47,6 +47,22 @@ ocp_worker_memory: 30000
...
+#ocp_bm_worker_vendor: Dell
+#ocp_bm_worker_bmc_address: 10.16.231.76
+#ocp_bm_worker_bmc_user: root
+#ocp_bm_worker_bmc_password: calvin
+#ocp_bm_worker_mac: 40:a6:b7:2b:19:01
+#ocp_bm_worker_ip: 192.168.111.30
+#ocp_bm_int: ens3f0
+
+# this address needs to be reachable from IPMI/iDRAC in case of BM worker 
+ocp_discovery_iso_server: 192.168.111.1
```
abays, line 63: One suggestion here: perhaps this is better suited for `vars/ocp_ai.yaml`, since it only pertains to AI.  We could also add `ai` into the name to make it more obvious (i.e. `ocp_ai_discovery_iso_server`).

ansible/ocp_ai_destroy.yaml
akiselev1: All bridges exist after successful completion of ocp_install. Looking at `make destroy_ocp`,  there is a  `TASK [Clean dev-scripts deployment]`, which does `make clean` for dev-scripts, which in turn kills bm and pr bridges.

ansible/ocp_ai_prep.yaml
akiselev1: I think we can use `when: ocp_bm_int is defined`, if nothing else.

ansible/vars/default.yaml
akiselev1: Done.

abays:
/lgtm

This will serve as basic functionality for a single baremetal OCP worker option, and later will can use it to bootstrap changes to support multiple baremetal OCP workers.

abays (approved):

PR #204, opened 2021-08-24
Add installing ansible collections from git

mpryc:
Use git if ansible-galaxy fails to find and
install collections.


abays (approved):
/lgtm

PR #213, opened 2021-09-02
Use ocp_version and ocp_minor_version

mpryc:
Use of ocp_version and ocp_minor_version allows CI to easily
use different ocp releases without putting too much logic in
the CI job definition. Also use latest rhcos-live image from the
particular release.


ansible/vars/ocp_ai.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-ocp_ai_discovery_iso: https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.7/4.7.0/rhcos-live.x86_64.iso
+ocp_ai_discovery_iso: "https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/{{ ocp_version }}/latest/rhcos-live.x86_64.iso"
```
abays, line 2: This will work for versions that have been publicly released, but it won't work for versions under-development (i.e. as of right now, 4.9).  We can address that later though.

stuggi (approved):
/lgtm

PR #215, opened 2021-09-03
Enable fencing only on 17.0 and add apiversion parameter

stuggi:
only enable fence_kubevirt in 17.0 environments, also add
apiserver support which got added to the fencing agent via:

https://github.com/ClusterLabs/fence-agents/pull/433


abays (approved):
/lgtm

PR #217, opened 2021-09-08
Allow to remove discovery image files on remote host

mpryc:
Replace with_fileglob matching, which is ran against
local system files on the Ansible controller with
syntax that searches for the files on the remote controller.

Current logic worked when the same user was invoking cleanup
on the same host that ansible playbook were stored.

### DIFF
--- a/ansible/ocp_ai_destroy.yaml
+++ b/ansible/ocp_ai_destroy.yaml
@@ -1,6 +1,7 @@
 ---
 - hosts: convergence_base
   become: true
+  become_user: root
   gather_facts: false
 
   tasks:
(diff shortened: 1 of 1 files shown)


ansible/ocp_ai_destroy.yaml
```diff
@@ -125,12 +126,18 @@
       path: "{{ osp_controller_base_image_url_path }}"
       state: absent
 
+  - name: Find discovery images
+    find:
+      paths: "/var/lib/libvirt/images/"
+      patterns: "discovery_image_ostest*.img"
```
abays, line 132: Could use `{{ ocp_cluster_name }}` here instead of hard-coding `ostest` cluster name

stuggi (approved):
/lgtm

PR #218, opened 2021-09-08
BM worker enhancements

akiselev1:
1. Use ansible module for BM interface creation and tear down. 
2. Re-order inventory.ospd.j2 for BM worker so it generates inventory correctly with the latest dependencies.
3. Add HTTP proxy config.

### DIFF
--- a/ansible/vars/default.yaml
+++ b/ansible/vars/default.yaml
@@ -241,3 +241,8 @@ virt_sriov_domains:
   - "{{ ocp_cluster_name }}_worker_4"
 
 kubevirt_fencing_agent_url: https://raw.githubusercontent.com/ClusterLabs/fence-agents/master/agents/kubevirt/fence_kubevirt.py
+
+# HTTP Proxy
+http_proxy: ""
+https_proxy: ""
+no_proxy: ""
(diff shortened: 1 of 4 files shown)


ansible/vars/default.yaml
```diff
@@ -241,3 +241,8 @@ virt_sriov_domains:
   - "{{ ocp_cluster_name }}_worker_4"
 
 kubevirt_fencing_agent_url: https://raw.githubusercontent.com/ClusterLabs/fence-agents/master/agents/kubevirt/fence_kubevirt.py
+
+# HTTP Proxy
+http_proxy: ""
+https_proxy: ""
+no_proxy: ""
```
abays, line 248: Could we prepend these with `ocp_ai_` and move them to `vars/ocp_ai.yaml`?  That way they get a clear association with the AI route of installation.

akiselev1: http_proxy is not specific to AI. It can be used for IPI as well, for example: https://openshift-kni.github.io/baremetal-deploy/4.8/Deployment.html#ipi-install-setting-proxy-settings-within-install-config_ipi-install-configuration-files

akiselev1: Discussed with @abays. We agreed to leave proxy settings in default.yaml for now for possible future use in IPI.

abays (approved):
/lgtm

PR #221, opened 2021-09-17
Allign dev-scripts and AI VM template

stuggi:
Make sure the AI VM template matches the one created by dev-scripts:
* provisioning network interface is ens1p0
* external network interface is enp6s0
* ospnetwork interface is enp7s0

### DIFF
--- a/ansible/templates/ai/libvirt/baremetalvm.xml.j2
+++ b/ansible/templates/ai/libvirt/baremetalvm.xml.j2
@@ -1,0 +1,112 @@
+<domain type='kvm'>
+  <name>{{ ocp_cluster_name }}-{{ role }}-{{ item }}</name>
+  <memory unit='MiB'>{{ memory }}</memory>
+  <vcpu>{{ vcpu }}</vcpu>
+  <os>
+    <type arch='x86_64' machine='q35'>hvm</type>
+    <loader readonly='yes' secure='yes' type='pflash'>/usr/share/edk2/ovmf/OVMF_CODE.secboot.fd</loader>
+    <nvram>/var/lib/libvirt/qemu/nvram/{{ ocp_cluster_name }}-{{ role }}-{{ item }}_VARS.fd</nvram>
+    <boot dev='network'/>
+    <bootmenu enable='no'/>
+  </os>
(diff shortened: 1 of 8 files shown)


ansible/templates/ai/libvirt/baremetalvm.xml.j2
```diff
@@ -0,0 +1,112 @@
...
+    <input type='mouse' bus='ps2'/>
+    <graphics type='vnc' port='-1' autoport='yes'/>
+    <video>
+      <model type='cirrus' vram='9216' heads='1'/>
+    </video>
+    <memballoon model='virtio'>
+      <alias name='balloon0'/>
+      <address type='pci' domain='0x0000' bus='0x05' slot='0x00' function='0x0'/>
+    </memballoon>
+    <rng model='virtio'>
```
olliewalsh, line 106: The RNG was explicitly removed in the previous approach. Any idea why?

stuggi: to have a free PCI slot. now the rng device gets set to 8 so that we have 6/7 available for the NICs

stuggi: correct on the free slot @abays ?

abays: Yes, the RNG was removed to free a slot beforehand

olliewalsh: Ah, ok. That *probably* wasn't necessary. IIUC can place multiple devices on a single root port (bus), just need to use a different slot. I think we could even use a single slot with different "functions" to emulate a nic with multiple ports.

abays (approved):
Changes look good to me

olliewalsh (approved):

ansible/templates/ai/libvirt/baremetalvm.xml.j2
abays: Agreed, it probably wasn't.  It was just a quick hack to move and deal with bigger fish. :-)

PR #223, opened 2021-09-22
 OSPK8-352 Switch to OSP16.2 as default

stuggi:
Also removes low-memory-usage.yaml to speed up tempest test runs.

Depends-On: https://github.com/openstack-k8s-operators/osp-director-operator/pull/393


abays (approved):
/lgtm

ansible/vars/default.yaml
```diff
@@ -196,7 +196,7 @@ osp:
     external: enp6s0
     ospnetwork: enp7s0
   network_type: vlan
-  container_tag: 17.0_20210907.1
+  container_tag: 16.2_20210901.2
```
dprince, line 199: perhaps in another commit we can support the publicly available OSP images (not our operator... but the OSP ones). As is this is going to still pull from registry-proxy.engineering.redhat.com/rh-osbs so internal only... but its okay for now.

stuggi: yes, in addition to the 16.2/17.0 we can also provide var files for upstream versions, and then default to one of them. I'll create a jira task for that

stuggi:
CI fails because of missing fencing patch for:

~~~
2021-09-24T08:06:18.753Z	INFO	controllers.OpenStackControlPlane	Underlying OSVMSet controller hit an error: secrets "osp-director-operator-kubevirtfencing" not found
~~~

PR #225, opened 2021-09-24
Fix corrupt grubenv when customizing guest image

olliewalsh:
Use grub2-mkconfig to update the kernel args in the grub environment
instead of sed. The env is binary and grub reports that it's corrupt
when trying to set a boot marker.


abays (approved):
/lgtm

stuggi (approved):
/lgtm

PR #228, opened 2021-09-28
OSPK8-364 use j2 template for rendering osnet CRs

stuggi:
Right now we have multiple temple files for the OSP networks.
This PR move to have a single j2 template and now it is possible
to specify networks using vars.

Also to not have to copy the full osp dict from var/defaults.yaml to
the version specific var file, this switches to use osp_defaults
dict in var/defaults.yaml and allows to overwrite specific parts
from the dict using osp_release_defaults dict in the version
specific, but also osp_local in local-defaults.yaml vars file.
Where required the combined osp dict gets created in the
playbook where it is currently used.


ansible/vars/16.2.yaml
```diff
@@ -1,34 +1 @@
...
-  interface:
-    external: enp6s0
-    ospnetwork: enp7s0
-  network_type: vlan
-  container_tag: 16.2_20210901.2
-  ceph_image: rhceph
-  ceph_tag: 5-12
-  # OSP deployment timeout
-  deploy_timeout: 60m
+osp: {}
```
olliewalsh, line 1: Could this be renamed to `osp_release_defaults` and use `osp` for the combined fact?

ansible/datavolume.yaml
```diff
@@ -6,21 +6,24 @@
   - oc_local
 
   tasks:
+  - name: Set combined osp dict
+    set_fact:
+      osp_combined: "{{ osp_defaults | combine((osp | default({})), recursive=True) }}"
```
olliewalsh, line 11: Could we combine with an `osp_local` fact too that could be set in local-defaults?

ansible/templates/osp/ctlplane/osp-director_openstackcontrolplane.yaml.j2
```diff
@@ -10,21 +10,21 @@ spec:
...
         - internalapi
-  openStackClientStorageClass: {{ osp.controller.storage_class }}
+  openStackClientStorageClass: {{ osp_combined.controller.storage_class }}
   passwordSecret: userpassword
   virtualMachineRoles:
     controller:
       roleName: Controller
-      roleCount: {{ osp.controller.count }}
+      roleCount: {{ osp_combined.controller.count }}
       networks:
```
olliewalsh, line 19: Should this be templated too?

ansible/vars/16.2.yaml
stuggi: you mean osp_releae_defaults in defaults.yaml, osp_release in the version specific overwrites and then osp as the combined to be used by the templates. That makes sense and I'd not have to change all the templates :)

ansible/datavolume.yaml
stuggi: ack good idea!

ansible/templates/osp/ctlplane/osp-director_openstackcontrolplane.yaml.j2
stuggi: yes for flexibility that makes sense. I'll do this in a follow up when working on subnet support

ansible/vars/16.2.yaml
stuggi: done

ansible/datavolume.yaml
stuggi: done

olliewalsh (commented):
/lgtm

abays (approved):
/lgtm

PR #230, opened 2021-09-29
Kuttl tests to be invoked using Makefile

mpryc:
KUTTL tests are written in the osp-director-operator git
repository. In order to invoke them we need to check out
that repository and execute those from it.

The repository in the CI is checked by the other parts,
so one can define path to the repository using env variable
or point directly to the upstream version and PR number
in the variable file.


stuggi (commented):
just the typo. other then that looks good to me

abays (approved):
/lgtm

PR #233, opened 2021-10-01
Add option to select mirror for guest image and rcm tools

mpryc:
In the event download.devel is down we need to be able to
pick other mirrors, so the CI is not failing.


mpryc:
Not really important anymore.

PR #234, opened 2021-10-05
Add k8s api route to controller VMs

stuggi:
Adds nic1 (pod network) to the controller nic template with a
route to 172.30.0.1/32 via 10.0.2.1 . In the end the nic templates
for VMs should get rendered by the osp-director-operator.


abays (approved):
/lgtm

PR #235, opened 2021-10-05
Initialize a --bare git repo

dprince:
This repository is meant to be used remotely only. This should
resolve some subtle issues that could occur when pushing/pulling
to the non-bare repo otherwise.


stuggi (approved):
/lgtm

PR #240, opened 2021-10-07
Add k8s api route to controller VMs part2

stuggi:
This is a follow up for 2935e16b2b88ebe446d5a7987c7e4d5e92a48a5b
to cover OSP17 controller VM nic template. Also fixes the template
to correct apply the br-ex default route.


olliewalsh (approved):
/lgtm

PR #244, opened 2021-10-14
add datavolume_cleanup make target

stuggi:


ansible/datavolume_cleanup.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - oc_local
+
+  tasks:
+  - name: Set combined osp dict
+    set_fact:
+      osp: "{{ osp_defaults | combine((osp_release_defaults | default({})), recursive=True) | combine((osp_local | default({})), recursive=True) }}"
+
+  - name: Set path to RHEL base image for dev-scripts
+    set_fact:
+      osp_controller_base_image_url_path: "{{ base_path }}/ironic/html/images/{{ osp.controller.base_image_url | basename }}"
```
olliewalsh, line 15: Do we need to append the base_image_url basename when  the delete task is removing it

abays (approved):
/lgtm

PR #247, opened 2021-10-21
OSPK8-374 Remove VM tripleo nic templates

stuggi:
The vm templates for multi nics get now rendered by the operator.

Depends-On: https://github.com/openstack-k8s-operators/osp-director-operator/pull/419


abays (approved):
/lgtm

PR #254, opened 2021-11-04
Use proper user when deinstalling/installing golang

mpryc:
If ansible user is different then root then appropriate
lifting perms needs to happen for the system wide tasks.


stuggi (approved):
/lgtm

PR #258, opened 2021-11-12
Configure TLS everywhere

olliewalsh:
Requires patched tripleoclient container with ipa client installed.
Runs freeipa in a container and redirects DNS for the base domain to it.
Configures IdmSecret when TLS-e is enabled and sets TLSe endpoint map.
Configures tempest to trust the freeipa CA.
Increases the OSP deploy timeout to 90m.

Depends-On: https://github.com/openstack-k8s-operators/osp-director-operator/pull/437


ansible/templates/osp/compute/openstackbaremetalset.yaml.j2
```diff
@@ -7,7 +7,7 @@ spec:
   # How many nodes to provision
   count: {{ osp.bmset[_role].count }}
   # The image to install on the provisioned nodes
-  baseImageUrl: http://192.168.111.1/images/{{ osp.base_image_url | basename }}
+  baseImageUrl: http://192.168.111.1:8080/images/{{ osp.base_image_url | basename }}
```
abays, line 10: This creates an issue, unfortunately, for the Metal3 dev-scripts deployment option.  We went with 80 for AI to allow for universal image URLs in our playbooks, as this is what dev-scripts configures for their http store by default.  There may be a way to configure it in dev-scripts, but I have never looked into it.  We could put some conditional logic here and elsewhere (or detect the deployment type and set some variable to either 80 or 8080) to handle the discrepancy, of course.

abays, line 10: This creates an issue, unfortunately, for the Metal3 dev-scripts deployment option. We went with 80 for AI to allow for universal image URLs in our playbooks, as this is what dev-scripts configures for their http store by default. There may be a way to configure it in dev-scripts, but I have never looked into it. We could put some conditional logic here and elsewhere (or detect the deployment type and set some variable to either 80 or 8080) to handle the discrepancy, of course.

ansible/vars/17.0.yaml
```diff
@@ -1,9 +1,9 @@
 # Todo: switch to downstream tripleoclient container image when available
-openstackclient_image: quay.io/openstack-k8s-operators/rhosp17-openstack-tripleoclient:17.0_20210720.1_patched2
+openstackclient_image: quay.io/openstack-k8s-operators/rhosp17-openstack-tripleoclient:17.0_20211111 # Upstream current-tripleo from 11/11/2011
```
abays, line 2: I think you meant 11/11/2021 in the comment.  11/11/2011 is the release date of Elder Scrolls V: Skyrim. :-)

ansible/templates/osp/compute/openstackbaremetalset.yaml.j2
olliewalsh: > or detect the deployment type and set some variable to either 80 or 8080

oops, that's what I intended to do :-)

ansible/templates/osp/tripleo_heat_envs/common/tlse.yaml.j2
```diff
@@ -0,0 +1,25 @@
...
+  OS::TripleO::Services::HAProxyInternalTLS: deployment/haproxy/haproxy-internal-tls-certmonger.yaml
+  OS::TripleO::Services::IpaClient: deployment/ipa/ipaservices-baremetal-ansible.yaml
+  OS::TripleO::Services::TLSProxyBase: deployment/apache/apache-baremetal-puppet.yaml
+  OS::TripleO::ControllerServiceServerMetadataHook: extraconfig/nova_metadata/krb-service-principals/controller-role.yaml
+  OS::TripleO::Compute{% if "hci" in osp.extrafeatures %}HCI{% endif %}ServiceServerMetadataHook: extraconfig/nova_metadata/krb-service-principals/compute{% if "hci" in osp.extrafeatures %}hci{% endif %}-role.yaml
+
+{% else %}
+parameter_defaults:
+  EnableInternalTLS: False
+{% endif %}
```
stuggi, line 25: I think we could render this using the operator and auto add it to the render playbook script in case tls-e is enabled.
like with the OSPversion we might want to add a status info the the controlplane CR to show either tls-e is enabled or not. This could then be used by other controllers if needed

ansible/templates/osp/tripleo_heat_envs/train/tlse_endpoint_map.yaml.j2
```diff
@@ -0,0 +1,98 @@
+{% if osp.tlse|default(false)|bool %}
+parameter_defaults:
+  EndpointMap:
```
stuggi, line 3: same, could just include environments/ssl/tls-everywhere-endpoints-dns.yaml in the create playbook when tls-e is enabled?

ansible/templates/osp/tripleo_heat_envs/17.0/tlse_endpoint_map.yaml.j2
```diff
@@ -0,0 +1,81 @@
+{% if osp.tlse|default(false)|bool %}
```
stuggi, line 1: same could just include environments/ssl/tls-everywhere-endpoints-dns.yaml in the create playbook when tls-e is enabled?

ansible/templates/osp/tripleo_heat_envs/16.2/tlse_endpoint_map.yaml.j2
```diff
@@ -0,0 +1,98 @@
+{% if osp.tlse|default(false)|bool %}
+parameter_defaults:
```
stuggi, line 2: could just include environments/ssl/tls-everywhere-endpoints-dns.yaml in the create playbook when tls-e is enabled?

ansible/templates/osp/tripleo_heat_envs/common/tlse.yaml.j2
olliewalsh: Yes... eventually. First I plan to add support for including built-in t-h-t environment files. Once that's there then most of this will no longer be necessary, just need to include environments/ssl/enable-internal-tls.yaml. Could then render IdmServer/Domain etc.. using the operator.

ansible/templates/osp/tripleo_heat_envs/train/tlse_endpoint_map.yaml.j2
olliewalsh: Yes, I think it would be safe to automagically include this in the create playbook provided it's lower precedence than user provided params. I need to look at some corner cases first - e.g Trilio overrides the endpoint map AFAIK.

olliewalsh: ... but if/when providing a list of built-in t-h-t environment files to include is possible then maybe that's all we need?

ansible/freeipa.yaml
```diff
@@ -0,0 +1,127 @@
...
+          hostname: freeipa.test.metalkube.org
+          image: quay.io/freeipa/freeipa-server:centos-8
+          #FIXME: need this to workaround a bug in podman ansible module
+          stop_signal: 37
+          env:
+            IPA_SERVER_IP: 10.89.0.2
+          volume:
+            - "/opt/freeipa/data/:/data:Z"
+          publish:
+            - 192.168.111.1:443:443
```
stuggi, line 52: seems to be one :443 to much?

olliewalsh: first port is just optional when its the same port number

stuggi: ack, thanks for clarification

ansible/templates/freeipa/ipa-server-install-options.j2
```diff
@@ -0,0 +1,7 @@
+-U
+-r {{ base_domain_name|upper }}
+--no-ntp
+--ds-password=abcd1234
```
stuggi, line 4: should we add the pwd which is used in a couple of places to a var in the defautl var file as this is probably the place people will look for if they can not remember?

stuggi (commented):
lgtm, just one question if it makes sense to add the pwd to the vars file for a central point to check when one don't know/can not remember

stuggi (approved):
/lgtm

PR #262, opened 2021-11-17
[OSPK8-328/OSPK8-388] Multi subnet / routed networks support

stuggi:
Depends-On: https://github.com/openstack-k8s-operators/osp-director-operator/pull/442

### DIFF
--- a/ansible/templates/osp/netconfig/osnet.yaml.j2
+++ b/ansible/templates/osp/netconfig/osnet.yaml.j2
@@ -1,0 +1,78 @@
+apiVersion: osp-director.openstack.org/v1beta1
+kind: OpenStackNetConfig
+metadata:
+  name: openstacknetconfig
+spec:
+  attachConfigurations:
+{% for nodeConfName, nodeConfPolicy in osp.attach_configs.items() %}
+    {{ nodeConfName }}:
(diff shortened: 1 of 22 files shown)


ansible/templates/osp/netconfig/osnet.yaml.j2
```diff
@@ -0,0 +1,78 @@
...
+              options:
+                stp:
+                  enabled: false
+              port:
+              - name: {{ nodeConfPolicy.interface }}
+            description: Linux bridge with {{ nodeConfPolicy.interface }} as a port
+            name: {{ nodeConfPolicy.bridge_name }}
+            state: up
+            type: linux-bridge
+{% if nodeConfPolicy.mtu is defined %}
```
bogdando, line 24: could kustomize help here to reduce the jinja templating?

abays (commented):
Looking good so far

ansible/templates/osp/netconfig/osnet.yaml.j2
stuggi: it might help, but then we'd have to template the kustomize config to be used in our dev-tools, so not sure if that helps that much.

abays (approved):
/lgtm

PR #263, opened 2021-11-19
dev-tools CI fixes

olliewalsh:
CNV sub is intermittently timing out while an operator upgrade occurs.
Image pull failing as CA isn't trusted


stuggi (approved):
/lgtm

PR #264, opened 2021-11-24
OSPK8-404 Fix cidr_suffix extract + ipv6 multi subnet

stuggi:
The cidr_suffix was not correct extracted from the subnet cidr
to create the gatway interface on the host. This fixes it and
also adds a task to create IPv6 gateway interface if the subnet
is IPv6.

Also moves the role_data.yaml to be a template that we only have
one file per version to maintain. Before there were two as we
had a separate one for IPv6 to add IPv6 and IPv4 gateway for the
controller role.


ansible/install_networks.yaml
```diff
@@ -47,13 +47,15 @@
   # Notes:
   # * can not use nmcli module https://github.com/ansible/ansible/issues/48055
   # * don't apply these persistent, in case of a host reboot those need to be reapplied
-  - name: Create gateways on isolated vlan networks
+  - name: Create IPv4 gateways on isolated vlan networks
+    vars:
+      cidr_suffix: "{{ item.1.ipv4.cidr | regex_search('.*/(.+)', '\\1') | first }}"
     shell: |
       set -e
-      cidr_suffix=$(echo "{{ item.1.ipv4.cidr }}" | sed  's/.*\///')
```
olliewalsh, line 53: this wasn't working?

olliewalsh (approved):
/lgtm

ansible/install_networks.yaml
stuggi: no, it only worked because it was default /24, realized that it does not work when testing ipv6 which has /64

olliewalsh: ah, so item.1.ipv4.cidr vs item.1.ipv6.cidr

stuggi: no, the var assignment in the script did not work. I also tried with something really simple. so I changed that :)

PR #273, opened 2021-12-16
OCS integration for AI

abays:
Allows for the (optional) deployment of OCS for AI OCP clusters


ansible/vars/default.yaml
```diff
@@ -333,18 +333,29 @@ osp_rhel_subscription_repos:
   - rhceph-4-tools-for-rhel-8-x86_64-rpms
   - codeready-builder-for-rhel-8-x86_64-rpms
 
-# Local Storage Operator specific (only required if using local-storage operator)
+# Local Storage Operator specific (only required if using local-storage operator or OCS)
+# NOTE: It is highly recommended to leave these as-is if you are enabling OCS
 local_storage_domains:
```
stuggi, line 338: wondering, do we need to specify them separate, or can  we just use the ocp_num_workers and make sure in case of OCS the min is 3?

ansible/ocp_ai_prep.yaml
```diff
@@ -387,6 +387,28 @@
...
+        ocs_local_storage: true
+        domain: "{{ (item | replace('_','-')) }}"
+      with_items: "{{ local_storage_domains }}"
+
+    - name: Prepare {{ local_storage_domains | join(',') }} VM(s) for use as storage node(s) for AI
+      shell: |
+        virsh setmaxmem {{ domain }} {{ ocp_storage_memory }}M --config;
+        virsh setmem {{ domain }} {{ ocp_storage_memory }}M --config;
+        virsh setvcpus {{ domain }} {{ ocp_storage_vcpu }} --config --maximum;
+        virsh setvcpus {{ domain }} {{ ocp_storage_vcpu }} --current
```
stuggi, line 407: if we use ocp_num_workers for the storage domains, we could already set those in the "define worker vms" task above?

abays: I've been working on an approach that uses a `num` var instead of an explicit list.  Trying it out this morning.  Let's see if it seems more suitable

stuggi (approved):
/lgtm

PR #274, opened 2021-12-22
Support for new osdeploy CR

dprince:
Updates the osp_register_overcloud_nodes.yaml so it avoids
using tripleo-deploy.sh to generate an Ansible inventory.

Instead this playbook generates a simple inventory
by using /etc/hosts on the openstackclient pod.

Also, updates the osp_deploy playbook to support the new
osdeploy CR.

### DIFF
--- a/ansible/osp_deploy.yaml
+++ b/ansible/osp_deploy.yaml
@@ -53,6 +53,14 @@
       <<: *oc_env
     register: deploy_switch
 
+  - name: Delete deployment (if it exists)
+    shell: |
+      set -e
+      oc delete -n openstack osdeploy default
+    environment:
+      <<: *oc_env
+    when: deploy_switch.stdout | length > 0
+
   - name: Start deployment
     shell: |
       set -e
--- a/ansible/osp_deploy.yaml
+++ b/ansible/osp_deploy.yaml
@@ -8,26 +8,62 @@
   - name: Include variables
     include_vars: vars/default.yaml
 
-  - name: Set combined osp dict
+  - name: Set directory for generator yaml files
     set_fact:
-      osp: "{{ osp_defaults | combine((osp_release_defaults | default({})), recursive=True) | combine((osp_local | default({})), recursive=True) }}"
+      deploy_yamls_dir: "{{ working_yamls_dir }}/deploy"
 
-  - name: Accept rendered playbooks
-    shell: |
-      #!/bin/bash
-      set -e
-      oc rsh -n openstack openstackclient /home/cloud-admin/tripleo-deploy.sh -a
+  - debug:
+      msg: "yamls will be written to {{ deploy_yamls_dir }} locally"
+
+  - name: Clean yaml dir
+    file:
+      state: absent
+      path: "{{ deploy_yamls_dir }}/"
+
+  - name: Create yaml dir
+    file:
+      path: "{{ deploy_yamls_dir }}"
+      state: directory
+      mode: '0755'
+
+  - name: Lookup configVersion hash
+    shell: >
+      oc get -n openstack --sort-by {.metadata.creationTimestamp} osconfigversions -o json | jq -re .items[0].spec.hash
+    register: config_version_cmd
     environment: &oc_env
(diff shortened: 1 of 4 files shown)


ansible/osp_register_overcloud_nodes.yaml
```diff
@@ -98,24 +87,25 @@
...
   - name: run rhsm playbook via openstackclient
     shell: |
       #!/bin/bash
-      oc rsh -n openstack openstackclient ansible-playbook -i /home/cloud-admin/playbooks/tripleo-ansible/tripleo-ansible-inventory.yaml /home/cloud-admin/rhsm.yaml
+      oc rsh -n openstack openstackclient <<"EOF_RSH"
+      cat > /home/cloud-admin/rhsm-inventory.yaml <<EOF
+      overcloud:
+        hosts:
+      EOF
+      for X in $(cat /etc/hosts); do
```
stuggi, line 103: wondering, do we already have the entries in /etc/hosts at this point? iirc they are only there when the osp deployment got run, but we have to register the nodes before. we could get the information from the ctlplane osnet, but shouldn't the director operator create this inventory for the user, or do we expect the end user to created this inventory file manually to register the nodes?

stuggi (commented):
should add a Depends-On: https://github.com/openstack-k8s-operators/osp-director-operator/pull/478 to not merge without?

ansible/osp_register_overcloud_nodes.yaml
dprince: Yeah, I had it working as I was re-running against an already deployed stack. But you are correct in that this won't work from scratch (just hit it in my testing).

ansible/osp_register_overcloud_nodes.yaml
```diff
@@ -98,24 +87,29 @@
...
     shell: |
       #!/bin/bash
-      oc rsh -n openstack openstackclient ansible-playbook -i /home/cloud-admin/playbooks/tripleo-ansible/tripleo-ansible-inventory.yaml /home/cloud-admin/rhsm.yaml
+      oc rsh -n openstack openstackclient <<"EOF_RSH"
+      cat > /home/cloud-admin/rhsm-inventory.yaml <<EOF
+      overcloud:
+        hosts:
+      EOF
+      for X in {{ ctlplaneIPs.stdout }}; do
+        echo "    $X:" >> /home/cloud-admin/rhsm-inventory.yaml
```
abays, line 110: Isn't `hosts` a list here?  Should this be `- $X` instead?

olliewalsh: Not a list, it's nested hashes. E.g https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html#assigning-a-variable-to-one-machine-host-variables

ansible/osp_deploy.yaml
```diff
@@ -8,26 +8,62 @@
...
+  - name: Start deployment
     shell: |
-      #!/bin/bash
       set -e
-      oc rsh -n openstack openstackclient timeout {{ osp.deploy_timeout }} /home/cloud-admin/tripleo-deploy.sh -p > {{ working_log_dir }}/osp-deploy.log 2>&1
+      oc apply -n openstack -f "{{ deploy_yamls_dir }}/openstackdeploy.yaml"
     environment:
       <<: *oc_env
-    when: playbooks_accept.rc == 0
+    when: deploy_switch.stdout | length == 0
```
olliewalsh, line 62: would it be better to delete the existing deployment?

dprince:
I'm going to re-work all the change to osp-register hosts here to use the new dynamic inventory. So before landing this we'll need to land both the osp-director-operator patches.

ansible/osp_register_overcloud_nodes.yaml
dprince: All gone now!

ansible/osp_deploy.yaml
dprince: Sure. Works for me I added a block before this in the latest patch to do this.

PR #275, opened 2022-01-06
Fix olm_cleanup

dprince:
Add namespace option to the oc command when looking up the currentCSV.

This resolves an issue with olm cleanup that got broken in
9dd498c529240147898bae3c562377b39be0469b


abays (approved):
/lgtm

PR #276, opened 2022-01-12
Allow switching between TLS/non-TLS deployments

olliewalsh:
Patch the AI dnsmasq.conf to redirect to freeipa DNS when configuring
freeipa instead of doing this at ocp_installer prep. Allows switching
between TLS/non-TLS openstack deployments. E.g
    make openstack_cleanup freeipa_cleanup freeipa openstack

Also install skopeo and jq in the download_tools playbook as they are
now required to build the operator.

Use the built-in t-h-t environment files for TLSe.


ansible/freeipa.yaml
```diff
@@ -125,3 +125,17 @@
           state: reloaded
           enabled: yes
         when: not (ocp_ai|bool)
+
+      - name: Update AI dnsmasq to forward to freeipa
+        lineinfile:
+          path: /etc/dnsmasq.d/dnsmasq_ai.conf
+          line: 'server=/{{ base_domain_name }}/10.89.0.2 # Forward to FreeIPA'
```
stuggi, line 132: or render the dnsmasq template again and verify if the config changed to restart the service?

olliewalsh: need to do something similar for dev_scripts (the tasks above this) so for consistency I think lineinfile is better. Could only trigger the NM/dnsmasq restart when the lineinfile task is changed though.

stuggi: ack, was just thinking if it is not cleaner to modify the cfg only in a single way (template) instead of template + lineinfile

stuggi (approved):
/lgtm

stuggi (approved):
/lgtm

PR #277, opened 2022-01-13
Fix ansible-lint by pinning it to 5.3.2 version

mpryc:
Due to recent issue in the ansible-lint package we need to
bump version which we are using:
  https://github.com/ansible-community/ansible-lint/issues/1795


.ansible-lint
```diff
@@ -4,3 +4,19 @@ skip_list:
   - '306'
   - '602'
   - '701'
+
+warn_list:  # or 'skip_list' to silence them completely
+  - experimental  # all rules tagged as experimental
+  - unnamed-task  # All tasks should be named
+  - var-spacing  # Variables should have spaces before and after:  {{ var_name }}
```
olliewalsh, line 11: Should we just use `# noqa var-spacing` for any exceptions to this rule?

mpryc: @olliewalsh the best would be to fix them that's why I didn't put them in the skip list just converted those from errors to warnings, so they are visible during CI runs and can be fixed in the follow PRs e.g one at a time rather then bigger rework. The list currently consists of 37 of those warnings. Making them warnings allows to quickly address issue with ansible-lint problem ansible-community/ansible-lint#1795 which was the main purpose of this PR.

olliewalsh: Also what does the linter have against ignore_errors! Should we disable that check?

mpryc: @olliewalsh skipped ignore_errors, it was complaining with:
`ignore-errors: Use failed_when and specify error conditions instead of using ignore_errors
`

added noqa to the specific lines.

PR #278, opened 2022-01-18
Fix 3-combo-node OCP cluster deployment

abays:
Allows for deployment of 3-node (master/worker combo) OCP clusters.  Verified for AI only.  Recommended `local-defaults.yaml` are as follows:

```
ocp_num_workers: 0
ocp_master_memory: 80000
ocp_master_vcpu: 24
ocp_master_disk: 100
```

### DIFF
--- a/ansible/ocp_ai_prep.yaml
+++ b/ansible/ocp_ai_prep.yaml
@@ -279,6 +279,10 @@
     become: true
     become_user: root
     block:
+    - name: set fact for total workers
+      set_fact:
+        total_workers: "{{ ocp_num_workers|int + ocp_num_extra_workers|int }}"
+
     - name: Delete existing VMs and disk QCOW2s (if any)
       shell: |
         for i in $(virsh list | grep "{{ ocp_cluster_name }}-" | awk '{print $2}'); do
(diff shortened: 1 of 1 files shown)


ansible/ocp_ai_prep.yaml
```diff
@@ -345,7 +349,8 @@
         virsh vol-create-as 'default'
         '{{ ocp_cluster_name }}-worker-{{ item }}'.qcow2 '{{ ocp_worker_disk }}'G
         --format qcow2
-      with_sequence: start=0 end={{ ocp_num_workers+ocp_num_extra_workers-1 }}
+      with_sequence: start=0 end={{ total_workers|int-1 if (total_workers|int-1) > 0 else 0 }}
```
abays, line 352: These `if`s prevent Ansible from throwing a `negative stride` sequence error (`start` cannot be `>` `end`).  Even if a `when` would otherwise skip the whole task, Ansible _still_ evaluates the `with_sequence` expression and throws the error.

stuggi (approved):
/lgtm

ansible/ocp_ai_prep.yaml
olliewalsh: I don't recall hitting this a few months ago when tried a 3 node deployment. Was on Fedora 34 IIRC so I assume it's  fixed in newer ansible versions.

olliewalsh (approved):

PR #280, opened 2022-01-19
Add support for TLS public endpoints

olliewalsh:
Always create caConfigMap. Defaults to just the Red Hat CA cert.

When public TLS is enabled (default), create a local CA and certs, add
the CA cert to caConfigMap.

Adds correctly logic in custom heat environments to support public_tls
only, tlse only, and public_tls with tlse for internal/admin endpoints.

Switch to modified tempest container image until patches land downstream.
Add Containerfile for the custom tempest container.


ansible/local_ca.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+      - name: Create local CA
+        command:
+          cmd: openssl req -x509 -nodes -newkey rsa:4096 -keyout /opt/local_CA/private/cakey.pem -out /opt/local_CA/cacert.pem -days 7300 -extensions v3_ca -subj '/C=US/ST=North Carolina/L=Raleigh/O=Red Hat/OU=OSP Director Operator devtools/CN={{ ansible_fqdn }}' -batch
+          creates: /opt/local_CA/cacert.pem
+
+      - name: Create server key
+        command:
+          cmd: openssl genrsa -out /opt/local_CA/private/openstack.pem 2048
+          creates: /opt/local_CA/private/openstack.pem
```
olliewalsh, line 58: need to chown this to the current user when not running as root

ansible/osp_tempest.yaml
```diff
@@ -88,8 +88,8 @@
       environment:
         <<: *oc_env
```
olliewalsh, line 90: TODO: add the overcloud public endpoint to the tempest /etc/hosts, or add it to dnsmasq

ansible/local_ca.yaml
```diff
@@ -0,0 +1,114 @@
...
+          creates: /opt/local_CA/certs/openstack/openstack.csr.pem
+
+      - name: Generate Certificate
+        command:
+          cmd: openssl ca -config /opt/local_CA/certs/openstack/openssl.cnf -in /opt/local_CA/certs/openstack/openstack.csr.pem -out /opt/local_CA/certs/openstack/openstack.crt.pem -batch -notext
+          creates: /opt/local_CA/certs/openstack/openstack.crt.pem
+
+      - name: Add OSP public endpoint to dev_script dnsmasq
+        lineinfile:
+          path: /etc/NetworkManager/dnsmasq.d/openshift-ostest.conf
```
abays, line 83: We could use `{{ ocp_cluster_name }}` here in of hard-coding `ostest`, in case the user changes the variable.

ansible/openstack_cleanup.yaml
```diff
@@ -44,3 +44,43 @@
...
+  - name: Remove local CA
+    become: true
+    become_user: root
+    file:
+      path: /opt/local_CA
+      state: absent
+
+  - name: Remove OSP public endpoint from dev_script dnsmasq
+    lineinfile:
+      path: /etc/NetworkManager/dnsmasq.d/openshift-ostest.conf
```
abays, line 57: Same with `ocp_cluster_name` here

ansible/local_ca.yaml
olliewalsh: well spotted!

ansible/templates/local_CA/openssl.cnf.j2
```diff
@@ -0,0 +1,378 @@
...
+ess_cert_id_chain       = no    # Must the ESS cert id chain be included?
+                                # (optional, default: no)
+ess_cert_id_alg         = sha256        # algorithm to compute certificate
+                                # identifier (optional, default: sha1)
+
+[ alt_names ]
+IP.1 = 192.168.25.100
+IP.2 = 10.0.0.10
+DNS.1 = 192.168.25.100
+DNS.2 = 10.0.0.10
```
stuggi, line 376: those IPs could be different if someone configures different networks. Would be better to get them from the osnet reservations, like `oc get osnet ctlplane -o json | jq -r .spec.roleReservations.ControlPlane.reservations[0].ip` and `oc get osnet external -o json | jq -r .spec.roleReservations.ControlPlane.reservations[0].ip`. we probably have to make sure we wait for the osctlplane to be ready.

ansible/Makefile
```diff
@@ -129,6 +136,7 @@ openstack:
 	$(MAKE) datavolume
 	$(MAKE) networks
 	$(MAKE) git
+	$(MAKE) local_ca
```
stuggi, line 139: see my bellow comment, we might want to move this after ctlplane and then can get the VIPs from the osnet information.

ansible/templates/local_CA/openssl.cnf.j2
olliewalsh: ack, right now they don't matter as the endpoints use the DNS names but I'll add support for IP endpoints while I'm fixing this

ansible/vars/default.yaml
```diff
@@ -273,6 +273,10 @@ osp_defaults:
   extrafeatures: []
   # Enable TLS everywhere
   tlse: false
+  # Enable TLS public_endpoints
+  tls_public_endpoints: true
```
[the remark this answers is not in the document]
stuggi: from our table regular jobs use 16.2_novacontrol_hci_tlse_subnet.yaml and 16.2_ipv6_tlse_subnet.yaml for 16. We probably should enable it in one of them I'd say.

PR #281, opened 2022-01-20
Reduce ingress controller replicas to 2

abays:
We have been defaulting to 3 ingress controller replicas due to an earlier issue with AI deployments.  This is no longer required.  It also was preventing upgrades of 3-combo-node OCP clusters from updating the MachineConfig operator.


olliewalsh (approved):
/lgtm

stuggi (approved):
/lgtm

PR #285, opened 2022-01-26
[OSPK8-431] Remove ext network from computes and relocate preserveReservations param

stuggi:
* Remove external network from compute nodes
Compute nodes do not need an IP on the external network,
therefore remove the external network form the role networks list
so that no IP get created.

* [OSPK8-431] Move preserveReservations parameter on level higher
This parameter now controls both MAC and IP reservations.

Depends-On: openstack-k8s-operators/osp-director-operator#495

### DIFF
--- a/ansible/openstack_cleanup.yaml
+++ b/ansible/openstack_cleanup.yaml
@@ -16,9 +16,11 @@
       - "oc delete -n openstack openstackephemeralheat --all"
       - "oc delete -n openstack openstackbaremetalset --all"
       - "oc delete -n openstack openstackcontrolplane --all"
+      - "oc delete -n openstack openstackvmset --all"
       - "oc delete -n openstack openstackclient --all"
       - "oc delete -n openstack openstacknetconfig --all"
       - "oc delete -n openstack cm tripleo-deploy-config heat-env-config tripleo-tarball-config --ignore-not-found=true"
+      - "oc delete -n opensatck events --all"
 
   - name: delete playbooks git repo dir
     become: true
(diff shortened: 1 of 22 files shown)


ansible/openstack_cleanup.yaml
```diff
@@ -16,9 +16,11 @@
       - "oc delete -n openstack openstackephemeralheat --all"
       - "oc delete -n openstack openstackbaremetalset --all"
       - "oc delete -n openstack openstackcontrolplane --all"
+      - "oc delete -n openstack openstackvmset --all"
       - "oc delete -n openstack openstackclient --all"
       - "oc delete -n openstack openstacknetconfig --all"
       - "oc delete -n openstack cm tripleo-deploy-config heat-env-config tripleo-tarball-config --ignore-not-found=true"
+      - "oc delete -n opensatck events --all"
```
abays, line 23: s/opensatck/openstack

abays (approved):
One small comment, otherwise /lgtm

ansible/openstack_cleanup.yaml
stuggi: fixed thanks

ansible/vars/16.2_ipv6_hci_tlse_subnet.yaml
```diff
@@ -26,24 +26,24 @@ osp_release_defaults:
   bmset:
     Compute:
       count: 0
-    ComputeLeaf1:
+    ComputeHCILeaf1:
```
olliewalsh, line 29: Why change the role name here but not in the other HCI roles?

stuggi: because I missed them :)

stuggi: added a new commit to add those

olliewalsh (approved):

PR #287, opened 2022-02-01
On cleanup, remove bindings of PV's in Failed or Released state

stuggi:
Also increase shared PVs amount to 8 since now RWX is the shared
to be used for VMs.


abays (approved):
/lgtm

olliewalsh (approved):
/lgtm

PR #288, opened 2022-02-02
With PR #478 [1] tripleo-deploy.sh is now at /usr/local/bin/tripleo-deploy.sh

stuggi:
Also no longer has the parameter to accept the new version. We should
instead use the inventory script from [2]

Also adds check on which tripleo-deploy.sh version is there that it still
can be used in current 1.2.x branch

[1] https://github.com/openstack-k8s-operators/osp-director-operator/pull/478
[2] https://github.com/openstack-k8s-operators/osp-director-operator/pull/487


stuggi:
dupe

PR #291, opened 2022-02-10
Front-load custom registries for AI

abays:
Adds a new install-time manifest to AI OCP deployments so that custom registries are already configured when the cluster deploys, rather than spending extra time post-install to reconfigure


stuggi (approved):
/lgtm

olliewalsh (approved):

PR #293, opened 2022-02-14
Check for 'finished' and 'error' during OSP deploy

abays:
Prevents `make osp_deploy` from hanging until timeout if an error is encountered during OSP deployment

### DIFF
--- a/ansible/osp_deploy.yaml
+++ b/ansible/osp_deploy.yaml
@@ -71,7 +71,38 @@
 
   - name: wait for deployment to finish
     shell: |
+      #!/bin/bash
       set -e
-      oc wait -n openstack osdeploy default --for condition=Finished --timeout="{{ (default_timeout * 40)|int }}s"
+
+      CUR_TIME=$(date +%s)
+      END_TIME=$(($CUR_TIME+{{ (default_timeout * 40)|int }}))
+
+      while [ "$CUR_TIME" -le "$END_TIME" ]; 
+      do
+        STATUS=$(oc get osdeploy -n openstack default | grep -v NAME | tail -1 | awk {'print $2'})
+        STATUS_CODE=$?
+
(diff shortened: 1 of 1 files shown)


ansible/osp_deploy.yaml
```diff
@@ -71,7 +71,37 @@
...
+        then
+          echo non-zero \"oc get osdeploy\" status code: $STATUS_CODE
+        fi
+
+        if [ "$STATUS" == "Finished" ];
+        then
+          exit 0
+        elif [ "$STATUS" == "Error" ];
+        then
+          echo Deployment hit an error. Please see OCP osdeploy resource/logs
```
olliewalsh, line 95: could dump the deploy pod logs to ostest-working/logs so the CI jobs gather them

ansible/osp_deploy.yaml
```diff
@@ -71,7 +71,37 @@
...
   - name: wait for deployment to finish
     shell: |
+      #!/bin/bash
       set -e
-      oc wait -n openstack osdeploy default --for condition=Finished --timeout="{{ (default_timeout * 40)|int }}s"
+
+      START_TIME=$(date +%s)
+      CUR_TIME=$START_TIME
+
+      while [ "$CUR_TIME" -le "$(($START_TIME+{{ (default_timeout * 40)|int }}))" ];
```
olliewalsh, line 80: (nit) why not:
```
CUR_TIME=$(date +%s)
END_TIME=$(($CUR_TIME+{{ (default_timeout * 40)|int }}))

while [  "$CUR_TIME" -le "$END_TIME" ]; 
```

ansible/osp_deploy.yaml
abays: That's a good idea

olliewalsh (approved):
/lgtm

PR #296, opened 2022-02-15
use LIBGUESTFS_BACKEND=direct for virt-customize

stuggi:


abays:
I've had to use the same thing in the past with regards to `virt-customize`.  Just curious though, what is requiring us to need it here?

stuggi:
> I've had to use the same thing in the past with regards to `virt-customize`. Just curious though, what is requiring us to need it here?

after the upgrade to rhel8.5 libvirt connection to libvirtd fail on the dev-scripts deployments for the ocp user when using virt-customize. since we can customize the qcow direct using the env var this can be prevented. not sure yet why we do not see this in 100% of the test runs.

abays:
> > I've had to use the same thing in the past with regards to `virt-customize`. Just curious though, what is requiring us to need it here?
> 
> after the upgrade to rhel8.5 libvirt connection to libvirtd fail on the dev-scripts deployments for the ocp user when using virt-customize. since we can customize the qcow direct using the env var this can be prevented. not sure yet why we do not see this in 100% of the test runs.

Ack, understood

abays (approved):
/lgtm

PR #298, opened 2022-02-16
Add cnosp role to deploying container native ctlplane

stuggi:
Adds a role and Makefile targets to deploy container native openstack control plane using following steps

```
make ocp_install
make cnosp
```
Tested with AI based deployment.

Note: right now know issue that cinder-volume service won't come up as kolla tries to change scsid config permissions on a read-only file system.


abays (approved):
/lgtm

olliewalsh (approved):
/lgtm

PR #304, opened 2022-03-01
Use Crucible for AI deployments

abays:
New AI deployer.  Only supports `4.X` granularity for now (where `X` >= 6).

Will explore version fine-tuning and SNO option next.


ansible/ocp_ai_destroy.yaml
```diff
@@ -80,13 +82,12 @@
...
+  ### ASSISTED INSTALLER SERVICE CONTAINERS
 
-  - name: Stop and remove assisted installer service container
+  - name: Stop and remove assisted installer service containers
     shell: |
-      ./stop-assisted-service.sh
-    args:
-      chdir: "{{ base_path }}/assisted-service-onprem"
+      podman pod stop assisted-service;
+      podman pod rm assisted-service
```
stuggi, line 90: on one of the ci nodes I had a situation where I had to remove the pod with force for some reason I don't know. should we just always force removing it?

ansible/vars/ocp_ai.yaml
```diff
@@ -23,10 +18,6 @@ ocp_ai_prov_bridge_worker_mac_prefix: 3c:fd:fe:78:cd:1
...
 
-# ocp_ai_service_store_dir: defaults to "/opt/assisted-service"
-# ocp_ai_service_port: defaults to "8090"
-
-# for now pin the ai image to a more stable tag - lastest started to fail on Feb 14th
-ocp_ai_service_image: quay.io/ocpmetal/assisted-service:v2.0.11
-
 ocp_ai_libvirt_storage_dir: /var/lib/libvirt/images
+
+ocp_ai_cli_version: "99.0.202112021829.202103111306"
```
stuggi, line 23: from where do we get that information, or do we just have to check which are there from pip?

stuggi (commented):
just two comments, other then that looks good to me to go as the replacement

ansible/ocp_ai_destroy.yaml
abays: Sure, we can add that.  The only concern is that, at least in my experience, the `force` option with podman sometimes leaves ports active and/or doesn't fully remove container storage.  I'll add the force for now and we can deal with it later if something goes awry.

ansible/vars/ocp_ai.yaml
abays: The current state of Crucible uses AI service images with API `v1`.  The `latest` `aicli` that gets installed by `pip` assumes it is targeting a `v2` API.  The version set in this Ansible variable just ensures we `pip install` an `aicli` version that assumes a `v1` AI API.  There is perhaps a way to configure (or pass a flag) to the latest version of `aicli` to tell it to use the `v1` API, but this was just the easiest solution for now.

stuggi: thanks for the details, maybe good to add them to the vars file why we use that version for now

abays: Ack

stuggi (approved):
/lgtm

PR #305, opened 2022-03-09
Fix train_subnet networks definition and wait for osnetcfg finished

stuggi:
Sometimes the create of the osnetcfg and related resources take longer and
later tasks fail, e.g. get ControlPlane IP. Lets wait for the osnetcfg to have its resources created correct or fail early.

Also fixes the used networks definition for train_subnet scenario.

### DIFF
--- a/ansible/install_networks.yaml
+++ b/ansible/install_networks.yaml
@@ -41,13 +41,20 @@
   - name: Start networks
     shell: |
       set -e
-      oc apply -n openstack -f "{{ network_yaml_dir }}"
+      oc apply -n {{ namespace }} -f "{{ network_yaml_dir }}"
     environment: &oc_env
       PATH: "{{ oc_env_path }}"
       KUBECONFIG: "{{ kubeconfig }}"
     # Retry on failure, probably osnet does not exist yet
     retries: 15
     delay: 10
+
+  - name: wait for osnetcfg to have configured interfaces + networks
+    shell: |
+      set -e
+      oc wait osnetcfg --for condition=provisioned -n {{ namespace }} openstacknetconfig --timeout="{{ (default_timeout * 2)|int }}s"
+    environment:
+      <<: *oc_env
 
   # Notes:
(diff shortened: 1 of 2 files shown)


abays (approved):
/lgtm

ansible/local_cert_tasks.yaml
```diff
@@ -9,7 +9,7 @@
         oc get -n {{ namespace }} osnet ctlplane -o json | jq -r .spec.roleReservations.ControlPlane.reservations[0].ip 
       register: get_control_plane_ip_ctlplane
       until: get_control_plane_ip_ctlplane.stdout|trim|ipaddr != False
-      retries: 60
```
olliewalsh, line 12: This normally only takes a couple of retries. Any idea what is causing such a big delay?

stuggi: when I have seen this the config of the nncp took quiet long and the osnetcfg create took longer then usually.

olliewalsh (approved):
/lgtm

ansible/local_cert_tasks.yaml
stuggi: I think I have a better idea. we should probably wait for in the install_networks tasks on  the osnetcfg to have finished before proceed or fail early at that point if there is an issue. that'd make debugging probably easier.

ansible/install_networks.yaml
```diff
@@ -41,14 +41,21 @@
...
     environment: &oc_env
       PATH: "{{ oc_env_path }}"
       KUBECONFIG: "{{ kubeconfig }}"
     # Retry on failure, probably osnet does not exist yet
     retries: 15
     delay: 10
 
+  - name: wait for osnetcfg to have configured interfaces + networks
+    shell: |
+      set -e
```
olliewalsh, line 54: nit: could just use command: instead of shell + set -e

olliewalsh (approved):
/lgtm

PR #309, opened 2022-03-11
bump container tags and reference tripleoclient tag from it

stuggi:


olliewalsh (approved):
/lgtm

olliewalsh:
maybe someday we can automate this...

PR #310, opened 2022-03-14
Support secrets repo for AI

abays:
Missed this inventory logic conversion when changing from `cluster_mgnt_roles` to Crucible


stuggi (approved):
/lgtm - we should probably document the secrets_repo better in the readme at some point. It does not clearly mention what can be added there.

PR #316, opened 2022-03-16
[OSPK8-449] Automate compose/puddle information based on input file

stuggi:
Adds osp_release_auto_url var which can be used to point to the
compose container_image_prepare.yaml. From there information is
parsed on the RHEL relase and containers to use to deploy OSP.

With this it would be no longer required to manually bump container
tags when a new compos is available.

Per default osp_release_auto_url points to the OSP16.2 passed_phase2
compose url.

### DIFF
--- a/ansible/vars/17.0.yaml
+++ b/ansible/vars/17.0.yaml
@@ -1,5 +1,6 @@
 ---
-osp_release_auto_url: http://download.devel.redhat.com/rcm-guest/puddles/OpenStack/17.0-RHEL-8/passed_phase1/container_image_prepare.yaml
+osp_release_auto_version: 17.0-RHEL-8
+osp_release_auto_compose: passed_phase1
 ephemeral_heat:
   heat_api_image: quay.io/tripleowallaby/openstack-heat-api:current-tripleo
   heat_engine_image: quay.io/tripleowallaby/openstack-heat-engine:current-tripleo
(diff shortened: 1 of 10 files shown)


ansible/vars/17.0.yaml
```diff
@@ -1,10 +1,5 @@
 ---
-openstackclient_image: registry-proxy.engineering.redhat.com/rh-osbs/rhosp17-openstack-tripleoclient:{{ osp_release_defaults.container_tag }}
-osp_rhos_release_compose: passed_phase1
-osp_release_defaults:
-  release: 17.0
-  container_tag: 17.0_20220214.1
-  ceph_tag: 5-102
+osp_release_auto_url: http://download.devel.redhat.com/rcm-guest/puddles/OpenStack/17.0-RHEL-8/passed_phase1/container_image_prepare.yaml
```
stuggi, line 2: should this come from secrets_repo instead?

stuggi: but we also had openstackclient_image url in the vars

abays (commented):
Seems good to me, but perhaps someone else should second-glance it from a deeper OpenStack perspective

stuggi:
lets wait for https://github.com/openstack-k8s-operators/osp-director-dev-tools/pull/319 which has the ansible-lint and network fix.

PR #319, opened 2022-03-17
[OSPK8-451] Move ctlplane network to the provisioning network for default MTU

stuggi:
There are couple of issues right now which make CI tests to fail:
1) After initial BMH provisioning the computes nodes enp1s0 and
enp2s0 are still configured for dhcp. As a result a default
route via ostestbm gets configured and clashes with the OSP
network config to have the default GW on the ctlplane network:

~~~
[cloud-admin@compute-1 ~]$ ip r
default via 192.168.25.1 dev br-tenant
default via 192.168.111.1 dev enp2s0 proto dhcp metric 101
~~~

This changes the OSP NIC template for the computes to disable both
mentioned networks on enp1s0 and enp2s0 during OSP network config
to have a single default GW configured.

Also removes no longer used `multiple_nics_dvr.j2`

2)  The ospnetwork has a MTU of 9000 so that it is possible to also use/test
jumbo frames with the storage/tenant network. The downstream repo
service does not/stopped to respond to ICMP, therefore pmtu does not
work properly and installing packages after ooo network config got
applied fail on the computes nodes, where the external network with
its default gw is not configured.
    
This moves the ctlplane network to the metal3 provisioning nic
enp1s0 which has default 1500 MTU.

3) Adds a simple forwarding only DNS server using dnsmasq to
the ctlplane network. Per default forwards to external_dns
server list. If tlse is enabled also adds 10.89.0.2 for the
freeipa pod to forward requests for the base_domain_name
to.


ansible/files/osp/16.2/net-config-two-nic-vlan-compute_leaf2.yaml
```diff
@@ -171,6 +171,16 @@ resources:
           params:
             $network_config:
               network_config:
+              # Disable BMH provisioning interface
+              - type: interface
+                name: nic1
+               use_dhcp: false
```
abays, line 177: Indentation of this line might need to move one to the right?

abays (approved):
/lgtm

ansible/templates/dnsmasq/ctlplane.conf.j2
```diff
@@ -0,0 +1,18 @@
...
+no-resolv
+
+# Don't use /etc/hosts
+no-hosts
+
+# Recursive DNS
+{% for dns_server in external_dns %}
+server={{ dns_server }}
+{% if osp.tlse|default(false)|bool %}
+server=/{{ base_domain_name }}/10.89.0.2 # Forward to FreeIPA
```
olliewalsh, line 11: In freeipa.yaml/freeipa_cleanup.yaml I'm patching the dnsmasq conf so we can easily swtich between TLS/non-TLS deployment: e.g https://github.com/openstack-k8s-operators/osp-director-dev-tools/blob/master/ansible/freeipa.yaml#L115?

olliewalsh: Looks like this dnsmasq conf is added during the openstack make target so I guess switching between TLSe/non-TLS may still work fine. Should this dnsmasq conf get removed in openstack_cleanup?

stuggi: yes, moved this endpoint for both install methods to this dnsmasq instance. Added a task in the cleanup to remove it.

PR #323, opened 2022-03-31
[OSPK8-417] remove dns settings from ctlplane and osclient CRs

stuggi:
The get now added by the operator from the osnetcfg if not specified.

Depends-On: https://github.com/openstack-k8s-operators/osp-director-operator/pull/556


abays (approved):
/lgtm

PR #328, opened 2022-04-06
Use v1-friendly AI client lib by default

abays:
A missing piece that aligns `aicli`'s underlying client library to AI API v1


stuggi (approved):
/lgtm

dprince (approved):

PR #329, opened 2022-04-07
[cnosp] create openstack-proto and fix cnosp_csv_version var name

stuggi:
openstack-proto sa is required for current ovn operator. Create it in its resources.
Also fixes misspelled cnosp_csv_version var name.


abays (approved):
/lgtm

PR #333, opened 2022-04-11
Cleanup insecure registry with 'destroy_ocp' task

dprince:
This patch adds a playbook go cleanup the insecure_registries
setting added in fde72ec21c462bcd3ffb63869ee79e7359b14db4 which
seems to break subsequent reinstallations with error messages
like:

loading registries configuration: loading registries configuration
\"/etc/containers/registries.conf\": mixing sysregistry v1/v2 is not
supported

Additionally the patch updates the name of the Ansible variable to
be 'podman_insecure_registries' instead of 'podman_insecure_registies'.

### DIFF
--- a/ansible/ocp_installer_destroy.yaml
+++ b/ansible/ocp_installer_destroy.yaml
@@ -23,3 +23,6 @@
 
 - name: Cleanup libvirt resources
   import_playbook: libvirt_cleanup.yaml
+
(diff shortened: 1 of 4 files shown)


abays (approved):
/lgtm

stuggi (approved):
/lgtm

PR #338, opened 2022-04-26
Fix osp_release_auto

olliewalsh:
The osp_release_auto yaml can persist between runs. `make openstack_cleanup`
will remove it, but `make olm_cleanup` (which is often run after
openstack_cleanup) includes the local role and fetch it again.

For now force it to be downloaded every time to ensure it does the
"right thing".


stuggi (approved):
/lgtm

abays (approved):
/lgtm

PR #344, opened 2022-05-03
Fix BMH provisioning for 3-combo-node OCP clusters

abays:
For 3-combo-node virtualized OCP clusters, we need to move the Metal3 provisioning interface target from `enp1s0` to the control plane bridge.  We attach `enp1s0` to the control plane bridge on the workers, and since the workers are also the masters (and since Metal3 sits there) we must ensure Metal3 points to the interface that actually holds the provisioning/ctlplane IP.

Also removes the wait for Metal3 pod to restart, as this is indirectly accounted for in the course of our reconcile loops anyhow.

### DIFF
--- a/ansible/install_networks.yaml
+++ b/ansible/install_networks.yaml
@@ -56,6 +56,19 @@
     environment:
       <<: *oc_env
 
+  # For virtualized 3-combo-node deployments, switch Metal3 provisioning network to the OSP control plane bridge
+  - name: Adjust Metal3 provisioning network for OCP 3-combo-node scenario
+    when: ocp_worker_count|int < 1 and ocp_num_masters > 0
+    block:
+    - name: Get name of OSP control plane bridge on OCP nodes
+      shell: |
+        oc get osnet -n {{ namespace }} ctlplane -o json | jq -r '.spec.attachConfiguration'
+      register: osnetconfig_ctlplane_intf
+
+    - name: Patch Metal3 Provisioning CR with desired interface name
+      shell: |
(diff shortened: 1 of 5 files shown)


ansible/install_networks.yaml
```diff
@@ -56,6 +56,20 @@
...
+    when: ocp_worker_count|int < 1 and ocp_num_masters > 0
+    block:
+    - name: Get name of OSP control plane bridge on OCP nodes
+      shell: |
+        oc get osnet ctlplane -o json | jq -r '.spec.attachConfiguration'
+      register: osnetconfig_ctlplane_intf
+
+    - name: Patch Metal3 Provisioning CR with desired interface name
+      shell: |
+        oc patch provisioning provisioning-configuration --type='json' -p='[{"op": "replace", "path": "/spec/provisioningInterface", "value": "{{ osnetconfig_ctlplane_intf.stdout }}"}]'
```
stuggi, line 71: what happens on openstack_cleanup [1] when the bridge interface is no longer there? Do we have to revert, or is it ok that its not there for some time?
[1] https://github.com/openstack-k8s-operators/osp-director-dev-tools/blob/master/ansible/openstack_cleanup.yaml#L23

abays: That's a good point.  Someone might just want to use a "vanilla" cluster again after having installed OpenStack, I guess.  I will add the reversal to the cleanup.

abays: I have added the cleanup and accounted for both AI and dev-scripts as well

stuggi (approved):
/lgtm

PR #345, opened 2022-05-04
[OSPK8-456] Use cephadm, not ceph-ansible, if osp.release >= 17

fultonj:
Update Heat environment templates storage-backend.yaml.j2
to use cephadm parameters in place of ceph-ansible parameters
if osp.release >= 17.


stuggi (approved):
/lgtm

PR #346, opened 2022-05-09
Introduce external_ceph to osp.extrafeatures

fultonj:
If osp.extrafeatures contains external_ceph, then
information about an already existing Ceph cluster
should be added to the external_ceph_data map in
vars/17.0_external_ceph.yaml or similar. The config
will then contain an updated storage-backend.yaml
with THT directives for tripleo to configure OpenStack
to use that external Ceph cluster.


stuggi (approved):
/lgtm

PR #349, opened 2022-05-16
set csv_version: 1.2.3 for testing release

pinikomarov:
Set csv ver.1.2.2 to correspond with this v1.2.x branch
This is to have consistent testing and pinning with regard to the final version we give to clients

### DIFF
--- a/ansible/vars/default.yaml
+++ b/ansible/vars/default.yaml
@@ -114,7 +114,7 @@ watch_namespace: openstack,openshift-machine-api,openshift-sriov-network-operato
 # osp-director-operator image and tag to use
 director_operator_image: quay.io/openstack-k8s-operators/osp-director-operator
 # CSV version is used for both the CSV version and container image tag
-csv_version: v1.2.x-latest
+csv_version: 1.2.2
 
 #ocp_network_type: OVNKubernetes
 ocp_network_type: OpenShiftSDN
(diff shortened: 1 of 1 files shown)


ansible/vars/default.yaml
```diff
@@ -114,7 +114,7 @@ watch_namespace: openstack,openshift-machine-api,openshift-sriov-network-operato
 # osp-director-operator image and tag to use
 director_operator_image: quay.io/openstack-k8s-operators/osp-director-operator
 # CSV version is used for both the CSV version and container image tag
-csv_version: v1.2.x-latest
+csv_version: 1.2.2
```
stuggi, line 117: latest 1.2 release is 1.2.3 [1], which was also released downstream [2]

[1] https://github.com/openstack-k8s-operators/osp-director-operator/releases/tag/v1.2.3
[2] https://access.redhat.com/errata/RHSA-2022:2183

pinikomarov: ack

pinikomarov:
@stuggi thing is , will the v1.2.latest all ways point to the version we give to the client ? 
is this is suitable tag for testing ? or should we switch to a statuic tag and update when we release 
new versions to clients  ?

stuggi:
> @stuggi thing is , will the v1.2.latest all ways point to the version we give to the client ? is this is suitable tag for testing ? or should we switch to a statuic tag and update when we release new versions to clients ?

right now we only backport specific patches v1.2 branch which get into a downstream release. There might be some delay, but I think it should be ok to use v1.2.x-latest per default. you always can overwrite it via local-defaults.yaml

pinikomarov:
leaving with default v1.2.x-latest to correspond with the dev-tools's   v1.2.x branch

Issue #351, opened 2022-05-18
AI based deployment is not using ocp_network_type configured in local-defaults


pmannidi:
When deploying AI based OCP, the network plugin is always default to OpenShiftSDN even if we set OVNKubernetes.

PR #352, opened 2022-05-18
Use user configured network type when deploying using AI

pmannidi:
Refer to ocp_network_type for user configured value instead of default OpenShiftSDN plugin

Resolves: #351 

Signed-off-by: Purandhar Sairam Mannidi [pmannidi@redhat.com](mailto:pmannidi@redhat.com)


abays (approved):
/lgtm

PR #357, opened 2022-05-23
Set passed_phase2 as default compose and add workaround

stuggi:
Add workaround for [1] to disable ping_test_gateway_ips
until the BZ is fixed.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2087231


olliewalsh (approved):
/lgtm works for me

PR #360, opened 2022-05-24
Clean infraenvs before AI deploy

abays:
Infraenvs are added in AI v2 and are separate from the cluster objects.  We should delete old ones just like we do for clusters before attempting an AI deployment.


stuggi (approved):
/lgtm

PR #368, opened 2022-06-08
Use correct ntp server on internal network

olliewalsh:
Use the correct ntp server on internal network, pool.ntp.org is blocked


abays (approved):
/lgtm

PR #371, opened 2022-06-10
Fix AI OCP version selection

abays:


olliewalsh:
CI failed - timed out waiting for config generator in openstack deploy . Maybe this patch should set the default version to 4.8.14 which we have been using all this time and can bump the version in a follow up patch?

abays:
> CI failed - timed out waiting for config generator in openstack deploy . Maybe this patch should set the default version to 4.8.14 which we have been using all this time and can bump the version in a follow up patch?

It looks like it eventually succeeded on another attempt.

PR #375, opened 2022-06-10
Fixes for hybrid (virt/BM) envs

abays:
Changes needed to support hybrid environments, where we have a virtualized OCP cluster and BM OSP computes


abays:
Closing in favor of #382

PR #376, opened 2022-06-13
Use root for all ostest-working tasks

olliewalsh:


olliewalsh:
issue is in the CI jobs

PR #378, opened 2022-06-14
Hybrid env support option 2

abays:
Keeps external OSP bridge as a Libvirt network, but turns Libvirt OSP network into a Linux bridge


abays:
Closing in favor of #382

PR #382, opened 2022-06-23
Yet another hybrid approach

abays:
This seems to be the simplest approach to adding support for hybrid environments while not regressing classic virtual environments.  The basic approach is:

- Change `external`/`ospnetwork` Libvirt networks into Linux bridges
- Add masquerading to `libvirt` firewall zone
- Add new Ansible variables for attaching prov host physical interfaces to the Linux bridges (if user so desires)

NOTE: Be sure to do `make destroy_ocp` _before_ pulling these bits if you have previously deployed dev-tools in your environment


ansible/ocp_vm_setup_extra_nics.yaml
```diff
@@ -13,71 +13,101 @@
...
-        copy:
-          src:  "{{ item }}"
-          dest: "{{ working_dir }}/osp-network/"
-          mode: 0644
-        with_items:
-          - files/ospnetwork.xml
-          - files/externalnetwork.xml
+      - name: Create ospnetwork bridge
+        shell: |
+          /usr/bin/nmcli connection add ifname ospnetwork type bridge con-name ospnetwork stp no ipv4.method disabled ipv6.method disabled ethernet.mtu 9000
```
stuggi, line 32: should the mtu here be/match the mtu set in for br-osp ethernet type interface in the attach_configs [1]?

[1] https://github.com/openstack-k8s-operators/osp-director-dev-tools/blob/master/ansible/vars/default.yaml#L356

ansible/ocp_vm_setup_extra_nics.yaml
```diff
@@ -13,71 +13,101 @@
...
+
+        - name: Check if ospnetwork interface is already attached to ospnetwork bridge
+          shell: |
+            /usr/bin/nmcli con show bridge-slave-{{ osp_bm_interface }}
+          ignore_errors: true
+          register: check_ospnetwork_bridge_exists
+
+        - name: Add ospnetwork interface to ospnetwork bridge
+          shell: |
+            /usr/bin/nmcli con add type bridge-slave ifname {{ osp_bm_interface }} master ospnetwork ethernet.mtu 9000
```
stuggi, line 52: same as above?

stuggi (commented):
looks good to me, just one question on the mtu for ospnetwork bridge

ansible/ocp_vm_setup_extra_nics.yaml
abays: Good call.  I added some logic to pull the MTU from the `attach_configs` data.

stuggi (commented):
/lgtm

stuggi (commented):
/lgtm

PR #386, opened 2022-06-28
[v1.2.x-backports] Yet another hybrid approach (#382)

abays:
Cherry-picks d1ba8387d9c40bd2208fa2390106c1ca51a9c8e5 and also one other small commit


abays:
Because CI won't work with this: ran successfully in my dev env

PR #394, opened 2022-07-21
Remove block to cleanup /etc/containers/registries.conf

stuggi:
This was kept for a while to cleanup previous way to add
custom registries. Lets remove it now.


abays (approved):
/lgtm

PR #396, opened 2022-08-01
Remove OS::TripleO::Services::Rear from roles

stuggi:
Rendering the playbooks fails with:

`2022-07-29T07:34:36.680892360Z heat.common.exception.StackValidationFailed: resources.ComputeHCILeaf1ServiceChain<file:///home/cloud-admin/tripleo-deploy-scratch/tripleo-heat-installer-templates/common/services/computehcileaf1-role.yaml>: The Resource Type (OS::TripleO::Services::Rear) could not be found.`

The service got removed from the roles in [1]. This also removes
it from the OSP17/wallaby roles file in the templates.

[1] https://github.com/openstack/tripleo-heat-templates/commit/ee6b260b85950899260493dae475ca85f7ad3825


abays (approved):
/lgtm

PR #398, opened 2022-08-05
use ctlplane_host_routes for ctlplane interfarce routes

stuggi:


ansible/files/osp/17.0/multiple_nics_vlans_dvr.j2
```diff
@@ -14,10 +14,7 @@ network_config:
   domain: {{ dns_search_domains }}
   addresses:
   - ip_netmask: {{ ctlplane_ip }}/{{ ctlplane_subnet_cidr }}
-  routes:
-  - ip_netmask: 0.0.0.0/0
-    next_hop: {{ ctlplane_gateway_ip }}
-    default: true
+  routes: {{ ctlplane_host_routes }}
```
abays, line 17: Where do we define this variable?  I can't seem to find it

stuggi: this is not a dev-tools template, this is an osp17 nic template which gets copied to the configgenerator to create the playbooks and run them. so those vars come from ooo

abays: Ah, right.

abays (approved):
/lgtm

PR #400, opened 2022-08-05
Fix deployed network environment ipv6

stuggi:
* add workaround for OSP17 until ipv6 for deployed-network-environment.j2.yaml is fixed
* make sure ip forwarding is enabled for ipv4 and ipv6
* remove workaround for https://review.opendev.org/c/openstack/tripleo-heat-templates/+/832520
* Also set ip6 GW address on the ext bridge to allow switch deployments

Depends-On: https://github.com/openstack-k8s-operators/osp-director-operator/pull/612

### DIFF
--- a/ansible/files/osp/17.0/environments/deployed-network-environment.j2.yaml
+++ b/ansible/files/osp/17.0/environments/deployed-network-environment.j2.yaml
@@ -36,11 +36,18 @@ parameter_defaults:
         subnets:
         {%- for subnet in network.subnets %}
           {{ subnet }}:
+          {%- if not network.ipv6 %}
             cidr: {{ network.subnets[subnet].ip_subnet }}
-            dns_nameservers: []
             gateway_ip: {{ network.subnets[subnet].gateway_ip }}
             host_routes: {{ network.subnets[subnet].routes | default([]) }}
             ip_version: 4
+          {%- else %}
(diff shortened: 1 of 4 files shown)


stuggi:
needs tempest update for ipv6 as we configured there the external network to be ipv6 only...

olliewalsh:
> * for ipv6 isolated networks, switch only external network to use ipv4

Why is this necessary?

stuggi:
> > * for ipv6 isolated networks, switch only external network to use ipv4
> 
> Why is this necessary?
> 
> > needs tempest update for ipv6 as we configured there the external network to be ipv6 only...
> 
> Ah, this? So we need a newer tempest image?

no, there was an issue connecting to internal registry/repos when we use ipv6 for the external network where the default gw points to. but now that we use the role nic templates it seem to work fine as we also set a ipv4 default gw via the ctlplane network. I'll update this PR as we need to set the external network GW on the host, like we do  with ipv4 here [1]

[1] https://github.com/openstack-k8s-operators/osp-director-dev-tools/blob/master/ansible/ocp_vm_setup_extra_nics.yaml#L65-L75

stuggi:
successful job (#15) of DFG-ospk8s-osp-director-dev-tools-16.2_ipv6_tlse_subnet using PR400

abays (approved):
/lgtm

olliewalsh (approved):

PR #404, opened 2022-08-18
Make adjustments for phase2 tempest tests

stuggi:
* disable block_migration_for_live_migration for hci as it is shared storage
* fix tempest_disable_feature_dict in tempest script template
*  Change project_network_v6_cidr to 2001:db8:fd01::/48. Tempest need to create multiple IPv6 /64 subnets, therefore we need to set a bigger project_network_v6_cidr. This change updates it to use 2001:db8:fd01::/48.
* Remove some tempest settings and use the defaults
* Disable tempest tests to only keep them on the 4 profiles for which there are downstream jobs. Therefore disable on:
    - upstream train
    - 16.2_novacontrol_hci_subnet.yaml
    - 16.2_ipv6_hci_tlse_subnet.yaml
    - 16.2_ipv6_novacontrol_hci_tlse_subnet.yaml
* Enable phase2 tempest test for 17.0_ipv6_subnet.yaml 
* Tempest modifications for IPv6
    -  Disable tests which test fip or required fip to validate for ipv6. With IPv6 only on the external network there are no fips. Some tests test fip functionality, others use fip to validate that the instance got created ok. Those obviously fail and the global feature setting network-feature-enabled.floating_ips=false does not help. This disables those tests which are included otherwise in the phase2 test list.
    - Disable Tests which expect to be able to set static external IP for IPv6
    - Disable live migration test with trunk for IPv6


stuggi:
IPv4:
```
Ran: 1363 tests in 2889.6757 sec.
 - Passed: 1268
 - Skipped: 95
 - Expected Fail: 0
 - Unexpected Success: 0
 - Failed: 0
Sum of execute time for each test: 9446.0143 sec.
```

abays (approved):
/lgtm

PR #406, opened 2022-08-23
Expose arbitrary inclusion of pre-existing heat env files

abays:
Exposes the `OsConfigGenerator`'s `heatEnvs` field for use with dev-tools via an Ansible list variable

### DIFF
--- a/ansible/templates/osp/config_generator/openstackconfiggenerator.yaml.j2
+++ b/ansible/templates/osp/config_generator/openstackconfiggenerator.yaml.j2
@@ -23,6 +23,9 @@ spec:
     - services/haproxy-public-tls-certmonger.yaml
     - ssl/enable-internal-tls.yaml
 {%- endif %}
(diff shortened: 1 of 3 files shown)


ansible/templates/osp/config_generator/openstackconfiggenerator.yaml.j2
```diff
@@ -7,6 +7,9 @@ metadata:
 spec:
   gitSecret: git-secret
   heatEnvs:
+{%- for envfile in osp_extra_env_files|default([]) %}
```
olliewalsh, line 10: might be best to append to the list since order is significant (last environment file wins if the same params are set)

abays: Good call

olliewalsh (approved):
/lgtm

PR #409, opened 2022-08-30
Add support for Ansible-STS test framework

udis:
- Add new sts pod created from tempest image
- Add sts templates to the pod
- Run sts and get results

### DIFF
--- a/ansible/osp_sts.yaml
+++ b/ansible/osp_sts.yaml
@@ -231,6 +231,8 @@
         #!/bin/bash
         oc exec -n openstack sts -- bash -c "rm -rf "{{ sts_home_dir }}/openstack-pidone-qe""
         oc exec -n openstack sts -- bash -c "pushd "{{ sts_home_dir }}/" && git clone -c http.sslVerify=false {{ sts_repo_url }}"
+      environment:
+        <<: *oc_env
       ignore_errors: true
 # sts pod pre END
 
@@ -268,6 +270,8 @@
         #!/bin/bash
         set -e
         oc exec -n openstack sts -- bash -c "pushd "{{ sts_home_dir }}/" ; sudo -H -u cloud-admin bash -c 'sh junit_report.sh'"
+      environment:
+        <<: *oc_env
       register: sts_report
 
     - name: Copy sts junit xml to localhost
--- a/ansible/osp_sts.yaml
+++ b/ansible/osp_sts.yaml
@@ -1,0 +1,305 @@
+---
+- hosts: localhost
+  vars_files: vars/default.yaml
+  roles:
+  - oc_local
+
+  tasks:
+  - name: Set combined osp dict
+    set_fact:
+      osp: "{{ osp_defaults | combine((osp_release_defaults | default({})), recursive=True) | combine((osp_local | default({})), recursive=True) }}"
+
+  - name: Verify openstackclient pod is running
+    shell: |
+        oc wait pod -n openstack openstackclient --for condition=Ready --timeout="{{ tempest_timeout }}s"
+    environment: &oc_env
+      PATH: "{{ oc_env_path }}"
+      KUBECONFIG: "{{ kubeconfig }}"
+    retries: 12
+    register: osp_client_pod_result
+    until: osp_client_pod_result.rc == 0
+
+# localhost prep START
+  - name: prepare localhost
+    when: osp_client_pod_result.rc == 0
+    block:
+    - name: Set sts yaml facts
+      set_fact:
+        sts_yaml_dir: "{{ working_yamls_dir }}/sts"
+        sts_working_dir: "{{ working_dir }}/sts"
+
+    - name: show yaml output dir
+      debug:
+        msg:
+        - "sts yamls will be written to {{ sts_yaml_dir }} locally"
+        - "sts config will be written to {{ sts_working_dir }} locally"
+
+    - name: Clean directories
+      file:
+        state: absent
+        path: "{{ item }}/"
+      loop:
+      - "{{ sts_yaml_dir }}"
+      - "{{ sts_working_dir }}"
+
+    - name: Create sts directories
+      file:
+        path: "{{ item }}"
+        state: directory
+        mode: '0755'
+      loop:
+      - "{{ sts_yaml_dir }}"
+      - "{{ sts_working_dir }}"
+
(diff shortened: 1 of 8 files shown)


ansible/osp_sts.yaml
```diff
@@ -0,0 +1,309 @@
...
+
+    - name: get IP per osp controller
+      shell: |
+         grep {{ item }}.ctlplane {{ sts_working_dir }}/openstackclient_hosts | cut -d " " -f1
+      loop: "{{ controller_name_list.stdout_lines }}"
+      register: controllers_ips
+
+    - name: get IP per osp compute
+      shell: |
+         grep {{ item }}.ctlplane {{ sts_working_dir }}/openstackclient_hosts | cut -d " " -f1
```
abays, line 89: Should `ctlplane` here be something else since we're greping for compute IPs?

abays: Nope, I was wrong.  This makes sense.

ansible/osp_sts.yaml
```diff
@@ -0,0 +1,309 @@
...
+          osp_computes: "{{ osp_computes | default([]) + [ {'name': item.item, 'ip': item.stdout} ] }}"
+      loop: "{{ computes_ips.results }}"
+
+    - name: set oc_nodes list
+      set_fact:
+        oc_nodes: "{{ osp_controllers|default([]) + osp_computes|default([]) }}"
+
+    - name: render sts osclient to {{ sts_yaml_dir }} dir
+      vars:
+        os_password: "{{ os_password_ret.stdout }}"
```
abays, line 109: I know we've used this pattern elsewhere, but where do we actually set `os_password_ret`?  I can't find it in dev-tools when I `grep -R os_password_ret *`.

udis: I used the osp_tempest.yaml as the basis for osp_sts.yaml - I guess that is why it's familier
I'll have a look at that parameter.

ansible/osp_sts.yaml
```diff
@@ -0,0 +1,309 @@
...
+    - name: Copy and edit ssh key file permission
+      shell: |
+        #!/bin/bash
+        set -e
+        oc exec -n openstack sts -- bash -c "sudo mv {{ sts_home_dir }}/id_rsa {{ sts_home_dir }}/.ssh/"
+        oc exec -n openstack sts -- bash -c "sudo mv {{ sts_home_dir }}/id_rsa.pub {{ sts_home_dir }}/.ssh/"
+        oc exec -n openstack sts -- bash -c "sudo chmod 600 {{ sts_home_dir }}/.ssh/id_rsa"
+        oc exec -n openstack sts -- bash -c "sudo chown cloud-admin:cloud-admin {{ sts_home_dir }}/.ssh/id_rsa"
+      environment:
+        <<: *oc_env
```
stuggi, line 227: wondering if we really need L22-L227 as this just replicates the openstackclient. why not just rsh to the openstackclient and do the bellow clone and run. the openstackclient already has a `/home/cloud-admin/ctlplane-ansible-inventory --list` ansible inventory script which can be used to easy connect to the overcloud nodes, run ansible commands ... it would make things a lot easier

ansible/osp_sts.yaml
```diff
@@ -0,0 +1,309 @@
...
+        set -e
+        oc exec -n openstack sts -- bash -c "cp -av {{ sts_operator_test_dir }}/* {{ sts_home_dir }}"
+      environment:
+        <<: *oc_env
+
+    - name: Verify ansible connectivty to overcloud nodes
+      shell: |
+        #!/bin/bash
+        set -e
+        oc exec -n openstack sts -- bash -c "pushd "{{ sts_home_dir }}/" ; sudo -H -u cloud-admin bash -c 'source ansible_ha.env && ansible controller -m shell -b -a 'hostname -s' '"
```
stuggi, line 254: could be the following if we use openstackclient:
~~~
oc exec -t -n openstack openstackclient -- /bin/bash -c "ansible -i ~/ctlplane-ansible-inventory -m ping Controller"
~~~

stuggi (commented):
I think we should try to use the openstackclient to run this . it has most of the requirements installed and we should probably be able just to checkout the sts repo and run it from there.

ansible/osp_sts.yaml
```diff
@@ -0,0 +1,309 @@
...
+# sts pod pre END
+
+# sts-ansible run tests and report START
+  - name: Run ansible-sts and generate report
+    block:
+    - name: Copy sts operator dir content
+      shell: |
+        #!/bin/bash
+        set -e
+        oc exec -n openstack sts -- bash -c "cp -av {{ sts_operator_test_dir }}/* {{ sts_home_dir }}"
```
stuggi, line 246: do we need to copy this? or can we just run from sts_operator_test_dir ?

ansible/osp_sts.yaml
udis: I was not able to find it anywhere... I can try to debug it and see if it's really needed.
I am wondering if the parameter is not defined ansible should have failed, right?

ansible/osp_sts.yaml
udis: Agree. This was the way things are tested manually. As in some cases, the openstackclient pod would be restarted due to OOM my thought was to dedicate a pod to it. I was able to use tempest pod image a quick w/a but it does require a lot of changes indeed.

I'll start the work on the openstackclient.

ansible/osp_sts.yaml
udis: ack

ansible/osp_sts.yaml
udis: We can leave it in the original path. It is just out of ease of access when debugging - nothing special.

udis:
Closing this and creating a new updated PR with all the changes

ansible/osp_sts.yaml
udis: Removed this task and will run from sts_operator_test_dir

ansible/osp_sts.yaml
udis: Used the suggested code from openstackclient

ansible/osp_sts.yaml
udis: PR 417 is running on the openstackclient and those line removed

PR #410, opened 2022-09-09
Fixes for DPDK testing

olliewalsh:
Fixes for DPDK testing.

Grubby messes with grub2 on RHEL8 when BLS is enabled (which is by default on the latest rhel8.4 kvm guest image).

Network is taking a very long time to come up on BM hosts so bump the timeout where we wait for this. Probably could improve some of the logic but I think this should be enough to get a successful DPDK deployment.


stuggi (approved):
/lgtm

PR #415, opened 2022-09-16
Workaround kernel cmdline insanity

olliewalsh:
Altering the kernel cmdline is completely different in RHEL8 and RHEL9 and AFAICT half-broken in both cases.

This logic seems to work on the RHEL and CentOS guest images we currently use.

### DIFF
--- a/ansible/datavolume.yaml
+++ b/ansible/datavolume.yaml
@@ -43,15 +43,41 @@
           mode: '0644'
           timeout: 30
 
-      - name: Remove net.ifnames=0 kernel param from {{ osp_base_image_url_path }}
-        shell: |
-          #!/bin/bash
-          set -e
-          virt-customize -a "{{ osp_base_image_url_path }}" \
-            --run-command 'sed -i -e "s/^\(GRUB_CMDLINE_LINUX=.*\)net.ifnames=0\s*\(.*\)/\1\2/" /etc/default/grub' \
-            --run-command 'grub2-mkconfig -o /etc/grub2.cfg' \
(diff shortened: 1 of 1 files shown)


ansible/datavolume.yaml
```diff
@@ -43,15 +43,41 @@
...
-          #!/bin/bash
-          set -e
-          virt-customize -a "{{ osp_base_image_url_path }}" \
-            --run-command 'sed -i -e "s/^\(GRUB_CMDLINE_LINUX=.*\)net.ifnames=0\s*\(.*\)/\1\2/" /etc/default/grub' \
-            --run-command 'grub2-mkconfig -o /etc/grub2.cfg' \
-            --run-command 'grub2-mkconfig -o /etc/grub2-efi.cfg' \
-            --run-command 'rm -f /etc/sysconfig/network-scripts/ifcfg-ens* /etc/sysconfig/network-scripts/ifcfg-eth*'
+      - name: Create virt-customize script to remove net.ifnames=0 kernel param from {{ osp_base_image_url_path }}
+        copy:
+          dest: "{{ osp_base_image_url_path }}"_customize.sh
```
pinikomarov, line 48: Hi @olliewalsh This should be : 
          dest: "{{ osp_base_image_url_path }}_customize.sh"
the **"**  quotes is what's breaking this and should be after the whole string

stuggi: works for me with the change Pini mentioned

pinikomarov:
lgtm

stuggi (approved):
/lgtm

PR #419, opened 2022-09-29
Disable network-feature-enabled.floating_ips for ipv6

stuggi:
As there is no ipv4 external network there are no floating IPs. Lets disable network-feature-enabled.floating_ips so that tests get skipped which require a floating ip and otherwise fail with

~~~
Details: {'type': 'BadRequest', 'message': 'Bad floatingip request: Network 0c2d2a84-0a4e-4cc7-9fd8-66eefa728424 does not contain any IPv4 subnet.', 'detail': ''}
~~~


abays (approved):
/lgtm

PR #420, opened 2022-10-02
Add tempest phase2 tests to 16.2.2_fencing.yaml

udis:


pinikomarov:
@udis what's special about the _excludelist_ ? is that the standard we're using ?

udis:
> 

@pinikomarov nothing special it's the standard phase2 tempest api and scenario testing

stuggi:
this is used as base 16.2.2 deployment for update testing, right? do we want to run there the full phase2 tests or is it ok to just run them after the update? I think we discussed that a while back, but can not remember the result. to run the phase2 tests twice requires another 45min iirc.

stuggi:
@udis is this still relevant not that we have [16.2.3_fencing.yaml](https://github.com/openstack-k8s-operators/osp-director-dev-tools/blob/master/ansible/vars/16.2.3_fencing.yaml) ?

pinikomarov:
@udis Hi please remind us what is the importance of this patch ? is it crucial ? 
It adds a lot more "osp phase2" checks right ? but just for the 16.2.2 job ? 
If it's not crucial we can drop it for now...
Thanks

stuggi:
closing this for now, lets reopen it if it is still useful

PR #424, opened 2022-10-25
Use latest osconfigversion for osdeploy

stuggi:
osconfigversions get sorted by date in [1] and first element of the configversions used to get the hash for the osdeploy. While this works for a fresh deployment where there is only a single osconfigversion, when an update is performed, the latest osconfigversion is the last element of the array.

[1] https://github.com/openstack-k8s-operators/osp-director-dev-tools/blob/master/ansible/osp_deployment.yaml#L34

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,13 +1,11 @@
 approvers:
   - abays
   - dprince
-  - mpryc
   - stuggi
   - olliewalsh
 
 reviewers:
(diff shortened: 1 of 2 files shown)


olliewalsh (approved):
/lgtm

abays (approved):
/lgtm

PR #427, opened 2022-11-21
STF support for 16.2

abays:


ansible/vars/default.yaml
```diff
@@ -400,6 +400,10 @@ osp_defaults:
   #    - hci
   #    - ipv6
   extrafeatures: []
+  # Any extra IP-hostname combos needed for resolution on overcloud nodes
+  extrahosts: []
```
olliewalsh, line 404: might be good to keep the same name as the t-h-t param, i.e extrahostfileentries

olliewalsh (approved):
/lgtm

PR #430, opened 2022-11-28
set nfv templates to use customer branch for operator

pinikomarov:
1.3.latest


pinikomarov:
contuing here : https://github.com/openstack-k8s-operators/osp-director-dev-tools/pull/432

PR #431, opened 2022-11-29
[OSPK8-647] Clean OSBMS first during 'make openstack_cleanup'

abays:
During `make openstack_cleanup`, we should delete any `OSBaremetalSet`s first and wait for their corresponding BMHs to deprovision before removing the rest of the OpenStack CRs and resetting the provisioning interface for Metal3.  Not honoring this order can occasionally interrupt the Metal3 deprovisioning process and cause the BMHs to get stuck in that state.

### DIFF
--- a/ansible/openstack_cleanup.yaml
+++ b/ansible/openstack_cleanup.yaml
@@ -5,18 +5,40 @@
   - oc_local
 
   tasks:
-  - name: openstack cleanup
-    command: "{{ item }}"
+  - name: BaremetalSet cleanup
+    shell: |
+      set -e
+      oc delete -n openstack openstackbaremetalset --all
+      sleep 10
     environment: &oc_env
       PATH: "{{ oc_env_path }}"
       KUBECONFIG: "{{ kubeconfig }}"
+    ignore_errors: true
+
+  - name: Wait for BMHs to settle
+    shell: |
+      for i in $(oc get bmh -A | grep -v STATE | grep openshift-worker | awk {'print $3'}); do
+          if [ "$i" != "available" ] && [ "$i" != "ready" ]; then
+             exit 1
+          fi
+      done
+    environment:
+      <<: *oc_env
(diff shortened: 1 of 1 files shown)


ansible/openstack_cleanup.yaml
```diff
@@ -5,18 +5,40 @@
...
+    shell: |
+      for i in $(oc get bmh -A | grep -v STATE | grep openshift-worker | awk {'print $3'}); do
+          if [ "$i" != "available" ] && [ "$i" != "ready" ]; then
+             exit 1
+          fi
+      done
+    environment:
+      <<: *oc_env
+    retries: "{{ (default_timeout / 2)|int }}"
+    delay: 5
```
stuggi, line 28: is this enough for the hybrid deployments with HW computes?

abays: Hmm, good point.  I guess since `default_timeout` is used all over the place, we couldn't just expect them to increase it to handle just this case.  Maybe we either need a new timeout variable for provisioning concerns or I can just remove the divided-by-2 here and hope that covers it?

stuggi: if we remove divided-by-2 it should be 20mins the. sounds ok. lets start with it and increase if needed

abays: Done

stuggi (approved):
/lgtm

olliewalsh (approved):
/lgtm

abays:
The CI failure here seems to be a 502 proxy error during `Download http://download.devel.redhat.com/rcm-guest/puddles/OpenStack/16.2-RHEL-8/passed_phase2/container_image_prepare.yaml` in the `olm` step.  It is thus not related to this change.

abays:
I tested this about a dozen times in the environment in which the problem was originally discovered, and the issue no longer occurred with this fix in place.

PR #433, opened 2022-12-07
Ignore errors from nmcli module when deleting bridges

olliewalsh:
This is failing on my dev host causing destroy_ocp to bail out early. Looks like an issue in the nmcli module or the nm command. For now just ignore errors.


stuggi (approved):
/lgtm

PR #436, opened 2022-12-08
Workaround BZ2087231 for 16.2.3

olliewalsh:
IIRC we previously had this workaround in dev tools but removed it when we switched to 16.2.4 composes. Re-enable it just for the 16.2.3 environment.


olliewalsh:
merging - 16.2.3 OSP deploy completed in QE job

stuggi (commented):
/lgtm right. I removed it when it passed with later phase2 compose. forgot about older releases :(

PR #437, opened 2022-12-09
Workarounds for sts

olliewalsh:


stuggi (approved):
/lgtm

ansible/osp_sts.yaml
```diff
@@ -100,7 +100,7 @@
       shell: |
         #!/bin/bash
         set -e
-        oc exec -t -n openstack openstackclient -- /bin/bash -c "source /home/cloud-admin/ansible_ha.env && pushd {{ sts_test_dir }} && timeout {{ osp.deploy_timeout }} sh osp_operator_sanity_run.sh" > {{ working_log_dir }}/sts.log 2>&1
+        oc exec -t -n openstack openstackclient -- /bin/bash -c "source /home/cloud-admin/ansible_ha.env && pushd {{ sts_test_dir }} && timeout {{ osp.deploy_timeout * 2 }} sh osp_operator_sanity_run.sh" > {{ working_log_dir }}/sts.log 2>&1
```
olliewalsh, line 103: doh, it's "90m" so can't just multiply it

ansible/osp_sts.yaml
```diff
@@ -86,6 +98,14 @@
       set_fact:
         sts_test_dir: /home/cloud-admin/openstack-pidone-qe/CI/ansible_sts_osp_operator
 
+    - name: Change instance boot script timeout to 10 minutes and hardcode external net
+      shell: |
+        #!/bin/bash
+        set -e
+        oc exec -n openstack openstackclient -- bash -c "sed -i -e 's/timeout_seconds=180/timeout_seconds=600/g' -e 's/EXTERNAL_NET_NAME=.*/EXTERNAL_NET_NAME=public/g' {{ sts_test_dir }}/templates/boot_oc_workload.sh.j2"
```
olliewalsh, line 105: Fetching the external net name was failing with a 503 Service Unavailable error, I expect it is just a timing issue. For now I've just hard-coded it

ansible/osp_sts.yaml
```diff
@@ -5,6 +5,18 @@
...
   tasks:
+
+  - name: Mark this is the first run
+    copy:
+      dest: "{{ working_dir }}/sts_marker"
+      content: ""
+      force: no
+    register: marker_status
+
+  - name: Skip if first run
```
olliewalsh, line 16: If we run sts before minor update it leaves a test db behind and the minor update fails, so skipping this step for now

pinikomarov (changes requested):
LGTM , and please drop the : 
tasks : 
  - name: Mark this is the first run
  - name: Skip if first run
  the fix for those was merged

PR #438, opened 2022-12-12
increase sts timeout

pinikomarov:


pinikomarov:
ready for reviews

stuggi:
two commits have the DNM == do not merge in the subject. should those be updated if we want to get that in?

dprince (approved):

pinikomarov:
that one was removed by me in the latest commit : https://github.com/openstack-k8s-operators/osp-director-dev-tools/pull/438/commits/e81542a03c0e23492515d22137089c8341fc42b4

and this one : https://github.com/openstack-k8s-operators/osp-director-dev-tools/pull/438/commits/56be06e615bb5ed1fe474231e0d630d5a15b7d7d
@udis do we need it ^^ ? 
the public net is set in the HA tests boot vm script no ? 
../CI/ansible_sts_osp_operator/templates/boot_oc_workload.sh.j2
EXTERNAL_NET_NAME=`openstack network list -c Name -f value --quote minimal |grep public || echo 'nova' `

pinikomarov:
This is the only change actually needed here : 
`sts_timeout|default('180m')` 
the increase boot time for the VM creation I can change in the downsream test repo on the test script itself

PR #442, opened 2023-01-05
truncate /etc/machine-id in golden image

stuggi:
virt-customize will create /etc/machine-id when customize the image used to provision the overcloud nodes. Make sure to use --truncate /etc/machine-id to clean /etc/machine-id so that a new machine-id gets created during provisioning


abays (approved):
/lgtm

PR #444, opened 2023-01-16
Allow pool_ids or pool regex to be used for rhsm

olliewalsh:
Allow either an explicit pool_id list or the existing pool regex to be used for subscription management.
Regex was not hitting the correct pools for me using the staging subscription server.


stuggi (approved):
/lgtm

PR #447, opened 2023-02-21
Update sushy-tools and crucible to latest versions

olliewalsh:
sushy-tools 0.20.0 fixes a caching issue which could result in requested
actions not occurring.

I expect this is the root cause of OCP installs failing intermittently while
waiting for a node to reboot.

0.20.0 and later require at least python 3.8 so jump to 3.9
Also run this in a virtualenv.

0.21.1 is broken so skip this version. I've landed a fix so should be in the next tag.

Latest commit on crucible repo is working with a minor change to the inventory. Now that the sushy-tools caching issue is fixed we can drop the workarounds.

### DIFF
--- a/ansible/host_prep.yaml
+++ b/ansible/host_prep.yaml
@@ -191,6 +191,10 @@
           # required to customize the guest-image to remove net.ifnames=0 kernel param
           - libguestfs-tools-c
 
+          # required to run recent sushy-tools
+          - python39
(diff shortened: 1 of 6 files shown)


olliewalsh:
Shaves 10mins off the deploy AFAICT

abays (approved):
/lgtm

PR #450, opened 2023-02-28
Fix OCP AI version selection

abays:
While our OCP AI version selection was previously working for ISO and root filesystem images, it was not properly finding and injecting the associated RHCOS version.  This fixes that.  It also removes our hacks for OCP 4.11/4.12 support and uses a more native approach as provided by Crucible.  

Furthermore, for the ISO and root filesystem images, we now use the latest available for the particular OCP version selected.  This is necessary because not all minor versions of a particular OCP version have associated ISO and root filesystem images available in the public OpenShift mirror.  I have done some spot-testing with the latest images with early versions of 4.10 and 4.11 (i.e. 4.10.2 and 4.11.2), and they seemed to work just fine.

### DIFF
--- a/ansible/ocp_ai_prep.yaml
+++ b/ansible/ocp_ai_prep.yaml
@@ -612,27 +612,13 @@
         repo: "{{ ocp_ai_ansible_repo | default('https://github.com/redhat-partner-solutions/crucible.git', true) }}"
         dest: "{{ base_path }}/crucible"
         force: true
-        version: "{{ ocp_ai_ansible_branch | default('863559ed9fdf61b650633922f4c29694b604ff09', true) }}"
+        version: "{{ ocp_ai_ansible_branch | default('3fc557583beba37fa3d084fae7c21410530f5cc6', true) }}"
 
     - name: Get BMC-network-connected interface's IP for baremetal deployments
       when: ocp_cluster_has_bm|bool
       shell: |
         /usr/bin/nmcli -g ip4.address device show {{ ocp_bmc_interface }} | cut -d '/' -f 1
       register: ocp_bmc_interface_ip
-
-    - name: Configure inventory variables
-      template:
-        src: ai/crucible/inventory.ospd.yml.j2
-        dest: "{{ base_path }}/crucible/inventory.ospd.yml"
-        mode: '0664'
-      vars:
-        ssh_pub_key: "{{ ssh_root_pub_key.stdout }}"
-
-    - name: Configure inventory vault variables
-      template:
-        src: ai/crucible/inventory.vault.ospd.yml.j2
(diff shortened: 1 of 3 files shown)


abays:
I have seen various operator subscriptions fail to install in the past day or two (CNV, SRIOV, NMState, etc).  Seems to be caused by authentication errors with the remote registry.  Then I try again and it works, so I think the CI failure here is a false indicator.

olliewalsh:
> I have seen various operator subscriptions fail to install in the past day or two (CNV, SRIOV, NMState, etc). Seems to be caused by authentication errors with the remote registry. Then I try again and it works, so I think the CI failure here is a false indicator.

ack, I've also hit problems over the past day or two, usually when pulling the OCP images though

ansible/ocp_ai_prep.yaml
```diff
@@ -657,42 +657,19 @@
         state: absent
         regexp: '^.*"schedulable_masters"'
 
-    # FIXME: Needed until upstream Crucible supports 4.11 and 4.12
-    - name: Inject Crucible support for OCP 4.11/4.12 provisioning images
-      when: ocp_version in ["4.11", "4.12"]
-      blockinfile:
+    - name: Override {{ ocp_version }} version ISO image
+      replace:
         path: "{{ base_path }}/crucible/roles/get_image_hash/defaults/main.yml"
```
olliewalsh, line 662: Could we just override these vars in the inventory instead of modifying the default vars in crucible?

ansible/ocp_ai_prep.yaml
```diff
@@ -711,7 +688,7 @@
       - name: Get RHCOS image version for {{ ocp_version }}.{{ ocp_minor_version }}
         set_fact:
           rhcos_image_version: "{{ rhcos_image_version_raw_data.content | regex_search('second_release=([^&]+)', '\\1') | default ([''], true) | first }}"
-          rhcos_image_version_alt: "{{ rhcos_image_version_raw_data.content | regex_search('\\/\\?release=([^&]+)', '\\1') | default ([''], true) | first }}"
+          rhcos_image_version_alt: "{{ rhcos_image_version_raw_data.content | regex_search('&amp;release=([^&]+)', '\\1') | default ([''], true) | first }}"
```
pinikomarov, line 691: +1 on This ,  Tested this and worked on a live env

ansible/ocp_ai_prep.yaml
abays: I will give that a try

abays: It does seem to work.  Will push changes to go that route.

stuggi (approved):
/lgtm

PR #464, opened 2023-04-25
Revert "4.12 master/worker workaround"

stuggi:
This reverts commit f6f6d1b2a259fd411a9b24817b96ee6f0f09a11e.

Now added to crucible.


abays (approved):
/lgtm

abays:
@stuggi We also need to update the version of Crucible we clone in dev-tools.  Right now it will be pinned to an old version that does not have the fix.

PR #468, opened 2023-05-04
Support enabling additional features for each custom deployment

olliewalsh:
Add a custom_config_extrafeatures var, defaults to an empty list, that is appended to the osp.extrafeatures list


krcmarik:
I've tested the PR successfully and used it as a base for PR #457

stuggi (approved):
/lgtm

PR #472, opened 2023-05-22
Copy the whole feature tarball config's directory content

krcmarik:
Similarly as in [1], copy the whole content of extrafeature tarball config directory and not only files (supposedly tht files) found in top directory. It's useful If someone wants to use patched tht files which will replaced the original tht files i.e. to workaround a bug in tht.

[1] https://github.com/openstack-k8s-operators/osp-director-dev-tools/blob/4462a78548928763a136c4cffa4fb7d225136171/ansible/osp_custom_config.yaml#LL84C1-L88C19


stuggi (approved):
/lgtm

PR #478, opened 2023-06-19
use preallocation=metadata for ceph disks

stuggi:


stuggi:
CI nodes disk is filling up when creating too many ceph nodes

krcmarik:
I've tested the patch successfully, +1 from me.

olliewalsh (approved):
/lgtm

PR #480, opened 2023-06-23
Trust local CA when running tempest

olliewalsh:
Previously would just warn that an unverified HTTPS request was being made, seems it's now stricter in 17.1


abays (approved):
/lgtm

PR #485, opened 2023-06-28
Make the PV NFS backend configurable

abays:
I use dev-tools for Podified and have found that more and more services require more PVs than the hard-coded defaults provide.  This allows users to define their PV backend quantity and sizes.


stuggi (approved):
/lgtm

PR #486, opened 2023-06-28
Dcn ci fixes

olliewalsh:
Change the subnet used for freeipa-network, it conflicts with a default podman network on newer systems.
Shorten the role name for DistributedComputeHCILeaf1 as it is used to generate the hostname and result in the gqdn exceeding a 64 char fqdn limit in the tripleo FreeIPA integration.
Default to TLSe for the DCN DMBS jobs.


stuggi (approved):
/lgtm

PR #489, opened 2023-07-03
add RH IT CA cert to host

stuggi:


abays (approved):
Tested in my env and it works

/lgtm

PR #496, opened 2023-07-12
Switch to the 0.5.2 cirros for DMBS deployments

krcmarik:
There is one tempest test [1] which seems to fail with 0.4.0 cirros image but passes with 0.5.2 cirros image on OSP 17.1.

[1] tempest.scenario.test_server_basic_ops.TestServerBasicOps.test_server_basic_ops


stuggi (approved):
/lgtm

PR #497, opened 2023-07-13
tempest glance img ephemeral

pinikomarov:
Allow tempest succesive runs - delete old glance images


pinikomarov:
Tested to work on dcn env 17.1 , tempest finished ok with no images left after the run

krcmarik:
I am wondering about the motivation? Do you need to run multiple tempest runs on DCN DMBS deployment? Is it for FFU? The image is created only when "dmbs" is specified in "feature" list. So I am just wondering.

pinikomarov:
Hi @krcmarik ,  it's for rerunning multiple tempests tests manually for a debug , for example when a yemtest tests fails and it's not a 100% reproducible issue

krcmarik (approved):

PR #501, opened 2023-07-28
Add support for 16.2->17.1 FFU

olliewalsh:
Add support for fast forward upgrade testing:
- fetch rhel8 compose info from the 17.1 rhel9 compose if available. Also cache the compose vars so we don't run this again for every imported playbook.
- reboot into new kernel after minor update to ensure it's running the new kernel before (leapp requires this)
- alternatively add option to update and reboot nodes after provisioning (default to this, much faster than running a minor update when we are already running latest 16.2)
- implement workflows for OSP FFU and HCI FFU and Multi-rhel FFU


ansible/osp_upgrade.yaml
```diff
@@ -0,0 +1,145 @@
...
+    deploy_advanced_settings:
+      limit: "nova_libvirt"
+
+# TODOs:
+# - debug why LVM backed cinder-volume did not come up after Leapp for single node controller
+# - run a canary workload, upgrade individual computes and live-migrate workload around, ping the canary VM throughout the FFU
+# - switch to ceph environment files instead of explicitly setting everying in storage_backend.yaml
+# - DCN dmbs
+# - external ceph?
+# - fix ceph osds on different subnets - storagemgmt routes are missing from netconfig templates, this is required for the heartbeat etc...
```
stuggi, line 145: I think thats something I fixed in https://github.com/openstack-k8s-operators/osp-director-dev-tools/commit/2d56cbda79be75cf7e779da7072fe4601e93d873 ?

ansible/osp_upgrade.yaml
```diff
@@ -0,0 +1,145 @@
...
+    deploy_mode: upgrade
+    deploy_advanced_settings:
+      limit: "nova_libvirt"
+
+# TODOs:
+# - debug why LVM backed cinder-volume did not come up after Leapp for single node controller
+# - run a canary workload, upgrade individual computes and live-migrate workload around, ping the canary VM throughout the FFU
+# - switch to ceph environment files instead of explicitly setting everying in storage_backend.yaml
+# - DCN dmbs
+# - external ceph?
```
stuggi, line 144: you mean with OSP17.1 we follow the traditional ceph deployment method?

stuggi: or testing with external ceph cluster?

ansible/templates/osp/tripleo_heat_envs/features/17.0/system_upgrade/system_upgrade.yaml.j2
```diff
@@ -0,0 +1,51 @@
...
+    crudini --get --format=lines leapp_upgrade_repositories.repo | grep ' enabled = 0' | awk '{print $2}' | xargs -i crudini --del leapp_upgrade_repositories.repo {}
+    crudini --del leapp_upgrade_repositories.repo rhos-release
+    crudini --del leapp_upgrade_repositories.repo rhos-release-brew
+    crudini --del leapp_upgrade_repositories.repo rhos-release-extras
+    sudo sed -i -e 's/\(.*\)enabled=1/\1enabled=1\n\1sslverify=0/g' leapp_upgrade_repositories.repo
+    # Put the upgrade repo leapp file in its proper place
+    cp leapp_upgrade_repositories.repo /etc/leapp/files/
+    
+    # TODO: FIXME: ACTOR vs THT
+    dnf -y remove irb
```
stuggi, line 48: could we add a comment why this has to be removed?

olliewalsh: no ours, this is from the regular FFU

ansible/osp_upgrade.yaml
olliewalsh: will remove that, was just a TODO from the beginning to check if anything was required for FFU with external ceph

ansible/templates/osp/tripleo_heat_envs/features/17.0/system_upgrade/system_upgrade.yaml.j2
```diff
@@ -0,0 +1,51 @@
...
+  # Ability to install leapp-repository package
+  LeappRepoInitCommand: |
+
+    rhos-release {{ osp_rhos_release }} -H {{ osp_rhos_release_mirror }} -r 8.4 -p {{ osp_release_auto_rhel8.compose }} -x
+
+  LeappInitCommand: |
+    # Fix broken grub on updated RHEL 8.4 guest image
+    echo 'GRUB_DEFAULT=saved' >> /etc/default/grub
+    grub2-mkconfig -o /etc/grub2.cfg
+    grub2-mkconfig -o /etc/grub2-efi.cfg
```
olliewalsh, line 22: This is the only change from the regular FFU config, to workaround a but on the RHEL 8.4 guest image

ansible/osp_upgrade.yaml
olliewalsh: ack, running a deploy to test now

ansible/osp_ceph_upgrade.yaml
```diff
@@ -0,0 +1,100 @@
...
+    custom_config_name: ceph-upgrade
+    config_generator_name: ceph-upgrade
+    config_generator_action: ceph-upgrade
+
+# Ceph 4->5 upgrade
+
+- name: Ceph Upgrade deployment
+  import_playbook: osp_deployment.yaml
+  vars:
+    ignore_deploy_error: "{{ workaround_BZ2226845|default(false) }}"
```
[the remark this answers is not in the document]
olliewalsh: from slack thread: will keep it for testing with specific versions if required later

ansible/osp_upgrade.yaml
olliewalsh: ceph upgrade passed so seems to be working now

PR #502, opened 2023-08-02
Move openstackclient FreeIPA cleanup to openstack_cleanup target

olliewalsh:
The openstackclient FreeIPA object cleanup was occurring in the wrong target - olm_cleanup. openstack_cleanup is the correct target (where the CR is removed).


abays (approved):
/lgtm

PR #503, opened 2023-08-04
Add proxy vars to configure OCP cluster wide proxy

stuggi:


ansible/vars/default.yaml
```diff
@@ -45,6 +45,11 @@ ocp_release_image: "quay.io/openshift-release-dev/ocp-release:{{ ocp_version }}.
 ocp_release_data_url: "https://amd64.ocp.releases.ci.openshift.org/releasestream/4-stable/release/{{ ocp_version }}.{{ ocp_minor_version }}"
 ocp_release_type: ga
 
+# OCP cluster wide proxy settings `oc get proxy/cluster`
+ocp_http_proxy: "http://squid.corp.redhat.com:3128"
+ocp_https_proxy: "http://squid.corp.redhat.com:3128"
+ocp_no_proxy: ".apps.ostest.{{ base_domain_name }},.usersys.redhat.com,192.168.111.0/24"
```
stuggi, line 51: not sure if we should set this in the job dev instead of here, thoughts?

olliewalsh: maybe leave the conf here but add a flag to enable it (disabled by default) that we can flip for individual jobs

ansible/templates/ai/crucible/inventory.ospd.yml.j2
```diff
@@ -38,6 +38,20 @@ all:
     # network_type: OpenShiftSDN
     network_type: {{ ocp_network_type }}
 
+    # Proxy settings. These settings apply to: Assisted Installer, Day1 clusters and Day2 clusters.
+    # This assumes the host where the AI runs and the OpenShift cluster share the same proxy settings.
+{% if enable_ocp_cluster_wide_proxy is defined and enable_ocp_cluster_wide_proxy is sameas true %}
```
olliewalsh, line 43: Never seen `is sameas true`. I assume that's equivalent to `var|bool == true`

olliewalsh: Could also have `var|default(false) is sameas true` to avoid the `is defined` check

olliewalsh (approved):
/lgtm

PR #506, opened 2023-08-11
Follow up to PR#502

stuggi:
[1] deleted the ansible/templates/freeipa/olm_cleanup.sh.j2 template which is still being called in this task. Lets remove it.

[1] https://github.com/openstack-k8s-operators/osp-director-dev-tools/pull/502


abays (approved):
/lgtm

PR #510, opened 2023-09-11
Improve overall OCP version support and add 4.13

abays:
Pins AI versions based on desired OCP version, which is needed to enable support for OCP 4.13.  Also adds general improvements for various version-related concerns.  These changes have been tested on:

- 4.10.63
- 4.11.48
- 4.12.13
- 4.13.8

Each version was tried as a 3-master-worker and a 3-master, 3-worker cluster.


stuggi (approved):
/lgtm

PR #514, opened 2023-11-22
Fix yaml indent when ceph_extraconfig_network is not defined

olliewalsh:
Ansible defaults to trims_blocks=True which messed up the indentation ceph_extraconfig_network is not set


stuggi (approved):
/lgtm

PR #516, opened 2023-12-07
stf templates

pinikomarov:
17.1 stf templates and tests execution

This add stf deployment options for 16.2 and 17.1 with added stf tests target in the makefile


pinikomarov:
Job passed this is ready for merge

ansible/osp_stf_tests.yaml
```diff
@@ -0,0 +1,51 @@
...
+    include_tasks: osp_sts_ha_tests.yaml
+    vars:
+        sts_script : 'osp_operator_workload_vm_run.sh'
+
+  # stf tests section
+  - name: Clone stf tests
+    shell: |
+      #!/bin/bash
+      oc exec -n openstack openstackclient -- bash -c "test -d /home/cloud-admin/functional-tests || (pushd /home/cloud-admin/ && git clone -c http.sslVerify=false https://github.com/infrawatch/functional-tests.git)"
+    environment: &oc_env
```
elfiesmelfie, line 18: I see you're using yaml anchors here.

You can also define the environment for the play up at the top, and it should propagate to all tasks, as long as the vars are defined (in this case, if they're in vars/default.yaml (from vars_files on L3), vars/${OSP_RELEASE}.yaml or local-defaults.yaml (from ansible-playbooks -e args)), then using environment for the play should work.

elfiesmelfie (approved):
The comment is more of a suggestion on how to improve maintainability -- I've been in the situation where I need env vars for shell commans, and kept forgetting to include them. Defining `environment` for the play will mean you don't need to do it per-task.

abays:
> The comment is more of a suggestion on how to improve maintainability -- I've been in the situation where I need env vars for shell commans, and kept forgetting to include them. Defining `environment` for the play will mean you don't need to do it per-task.

Ah, we weren't aware of that approach.  I agree that it could clean things up a bit.  Perhaps a follow-up across the repo later?  Otherwise the changes look good to me.

abays (approved):
/lgtm

pinikomarov:
@pinikomarov do we need anything else to get this merged ? after the lgtm this was supposed to be merged no ?

abays:
> @abays do we need anything else to get this merged ? after the lgtm this was supposed to be merged no ?

These have to be manually merged.  I will do it.

PR #520, opened 2024-01-04
stf templates update

pinikomarov:
update stf templates according to official docs : 
 https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/17.1/html-single/service_telemetry_framework_1.5/index
and IR templates : https://github.com/infrawatch/service-telemetry-operator/tree/master/tests/infrared/17.1

### DIFF
--- a/ansible/templates/osp/tripleo_heat_envs/features/17.0/stf/enable-stf.yaml.j2
+++ b/ansible/templates/osp/tripleo_heat_envs/features/17.0/stf/enable-stf.yaml.j2
@@ -1,46 +1,60 @@
 parameter_defaults:
-    CeilometerQdrPublishEvents: true
+    # only send to STF, not other publishers
+    PipelinePublishers: []
+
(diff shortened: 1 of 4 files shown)


ansible/templates/osp/tripleo_heat_envs/features/17.0/stf/enable-stf.yaml.j2
```diff
@@ -1,46 +1,60 @@
 parameter_defaults:
-    CeilometerQdrPublishEvents: true
+    # only send to STF, not other publishers
+    PipelinePublishers: []
```
yadneshk, line 3: Since Gnocchi resources are enabled in gnocchi-connectors-rdb.yaml.j2, I am assuming that you want to include Gnocchi in your deployment. In that case, please remove this.  Empty list results in no metric data passing to RHOSP telemetry components, such as Gnocchi or Panko.

mrunge:
This looks largely okay, while I want to note, that I did not test this.

PR #524, opened 2024-02-29
Enable octavia in 16.2_fencing and 17.1_fencing to be included in ffu

stuggi:


stuggi:
hit https://bugs.launchpad.net/tripleo/+bug/1907112 in the octavia 16.2 deployment ... seems this not have been backported to train. can not use octavia without a backport to train.

PR #525, opened 2024-02-29
Prototype of RHEL9 support

abays:
These changes allowed me to run `make default` successfully in a RHEL 9.3 environment.  I also tested some of them on RHEL 8, but not everything here has been tested there by me.  CI, however, passed on RHEL8 here [1].

To prepare the RHEL 9.3 environment, I simply used `subscription-manager` to register.  I did not attach to any specific pool.

[1] https://rhos-ci-jenkins.lab.eng.tlv2.redhat.com/blue/organizations/jenkins/DFG-ospk8s-osp-director-dev-tools-ai-pr/detail/DFG-ospk8s-osp-director-dev-tools-ai-pr/643/pipeline

### DIFF
--- a/ansible/install_networks.yaml
+++ b/ansible/install_networks.yaml
@@ -81,7 +81,8 @@
     become: true
     become_user: root
     shell: |
-      firewall-cmd --zone=trusted --add-source=0.0.0.0/32
+      firewall-cmd --zone=trusted --add-source=0.0.0.0/32 --permanent
+      firewall-cmd --reload
 
   - name: Create IPv4 gateways on ctlplane network
     become: true
--- a/ansible/install_networks.yaml
+++ b/ansible/install_networks.yaml
@@ -75,6 +75,14 @@
   # * can not use nmcli module https://github.com/ansible/ansible/issues/48055
   # * don't apply these persistent, in case of a host reboot those need to be reapplied
 
+  # REMOVEME?: Seems to be needed in RHEL 9, otherwise PXE boot requests coming in on
+  # ostestpr bridge from source 0.0.0.0 are dropped
+  - name: Allow BOOTP 0.0.0.0 source on trusted network
+    become: true
+    become_user: root
+    shell: |
+      firewall-cmd --zone=trusted --add-source=0.0.0.0/32
+
   - name: Create IPv4 gateways on ctlplane network
     become: true
     become_user: root
--- a/ansible/install_networks.yaml
+++ b/ansible/install_networks.yaml
@@ -81,8 +81,10 @@
     become: true
     become_user: root
     shell: |
-      firewall-cmd --zone=trusted --add-source=0.0.0.0/32 --permanent
-      firewall-cmd --reload
+      firewall-cmd --zone=trusted --add-source=0.0.0.0/32 {{ item }}
+    with_items:
+      - ""
+      - "--permanent"
 
   - name: Create IPv4 gateways on ctlplane network
     become: true
--- a/ansible/templates/ai/libvirt/baremetalvm.xml.j2
+++ b/ansible/templates/ai/libvirt/baremetalvm.xml.j2
@@ -4,8 +4,6 @@
   <vcpu>{{ vcpu }}</vcpu>
   <os>
     <type arch='x86_64' machine='q35'>hvm</type>
-    <loader readonly='yes' secure='yes' type='pflash'>/usr/share/edk2/ovmf/OVMF_CODE.secboot.fd</loader>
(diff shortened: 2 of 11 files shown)


ansible/vars/default.yaml
```diff
@@ -525,3 +525,87 @@ must_gather_image: quay.io/openstack-k8s-operators/must-gather:latest
...
+rhel_repos:
+  8:
+  - rhel-8-for-x86_64-appstream-rpms
+  - advanced-virt-for-rhel-8-x86_64-rpms
+  - openstack-16-for-rhel-8-x86_64-rpms
+  9:
+  - rhel-9-for-x86_64-appstream-rpms
+  - codeready-builder-for-rhel-9-x86_64-rpms
+
+rhel_packages:
```
abays, line 538: The distinction here might actually be unnecessary at this point.  I created this initially in a different RHEL 9 environment than the one I ultimately used to finish this first draft.  That old environment was cluttered with extra repo content from automated machine provisioning.

abays:
This passed CI here but for some reason is not reflected in the PR: https://rhos-ci-jenkins.lab.eng.tlv2.redhat.com/blue/organizations/jenkins/DFG-ospk8s-osp-director-dev-tools-ai-pr/detail/DFG-ospk8s-osp-director-dev-tools-ai-pr/643/pipeline

ansible/templates/ai/libvirt/baremetalvm.xml.j2
```diff
@@ -4,8 +4,6 @@
   <vcpu>{{ vcpu }}</vcpu>
   <os>
     <type arch='x86_64' machine='q35'>hvm</type>
-    <loader readonly='yes' secure='yes' type='pflash'>/usr/share/edk2/ovmf/OVMF_CODE.secboot.fd</loader>
```
olliewalsh, line 7: why drop this? Required for UEFI I believe

abays: Because it was so long ago, all I can remember was that for whatever reason this line was problematic and prevented the VMs from booting in RHEL 9 with whatever version of Libvirt it had.  We don't seem to use UEFI, but maybe we should be doing so (or perhaps I'm just missing where we in fact do).

ansible/vars/default.yaml
```diff
@@ -526,3 +526,87 @@ must_gather_image: quay.io/openstack-k8s-operators/must-gather:latest
...
+  - qemu-kvm
+  - virt-install
+
+  # required to customize the guest-image to remove net.ifnames=0 kernel param
+  - libguestfs-tools-c
+
+  # required to run recent sushy-tools
+  - python39
+  - python39-devel
+  - python3-virtualenv
```
dprince, line 574: Just ran this on RHEL 9.4 and I think we are missing 'virtualenv' there. We'll need to find the right source or replacement for this or change out the playbook works on RHEL9

abays: I couldn't find the equivalent `python3-virtualenv` package for RHEL 9.  And I think I remember having to manually install `virtualenv` via `pip`, as I think you noted in your comment here [1].

[1] https://github.com/openstack-k8s-operators/osp-director-dev-tools/pull/525#issuecomment-2375568179

ansible/install_networks.yaml
```diff
@@ -75,6 +75,14 @@
...
   # * don't apply these persistent, in case of a host reboot those need to be reapplied
 
+  # REMOVEME?: Seems to be needed in RHEL 9, otherwise PXE boot requests coming in on
+  # ostestpr bridge from source 0.0.0.0 are dropped
+  - name: Allow BOOTP 0.0.0.0 source on trusted network
+    become: true
+    become_user: root
+    shell: |
+      firewall-cmd --zone=trusted --add-source=0.0.0.0/32
+
```
dprince, line 85: The traffic I see being blocked looks like this:

Sep 26 11:38:07 osp-converge-03.lab.eng.brq2.redhat.com kernel: filter_IN_public_REJECT: IN=eno1 OUT= MAC=ff:ff:ff:ff:ff:ff:2c:76:8a:52:7b:61:08:00 SRC=0.0.0.0 DST=255.255.255.255 LEN=326 TOS=0x00 PREC=0x00 TTL=64 ID=0 DF PROTO=UDP SPT=68 DPT=67 LEN=306 

Also, if we are going to add a rule like this I think we'll want to add --permanent, and then do a 'firewall-cmd --reload'

ansible/vars/default.yaml
dprince: I grabbed a fix from ollie's branch for this. Should be all good with that

PR #527, opened 2024-04-12
bump ocp version to 4.14

stuggi:


stuggi:
close in favor https://github.com/openstack-k8s-operators/osp-director-dev-tools/pull/528

PR #530, opened 2024-04-19
bump ocp worker memory to 35G

stuggi:


stuggi:
osp-director-dev-tools-17.1_dcn_dmbs job was successful after bumping the workers mem

PR #534, opened 2024-05-13
increase shared 128gb RWX pvs to 10

stuggi:


stuggi:
all rwx pvs were consumed on a job and therefore failed

PR #547, opened 2024-10-28
default to 4.16

stuggi:


ansible/vars/default.yaml
```diff
@@ -40,7 +40,7 @@ podman_insecure_registries:
 
 # To set a specific release to install.
 # TODO: support "latest"/"stable" e.g https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable-4.14/
-ocp_version: "4.14"
+ocp_version: "4.16"
 ocp_minor_version: 19
```
abays, line 44: Is this the `4.16.X` version we want?  Not sure what's available.

stuggi: there is a `4.16.19` from where we pull the image, thats why I kept it

stuggi:
lets wait for all ci nodes to be migrated to rhel9

abays (approved):
All nodes are now RHEL 9

/lgtm

PR #548, opened 2024-11-12
Disable BMH disk cleaning by default in AI installations

abays:
It might not be necessary to always clean BMH disks during provisioning and deprovisioning.  Not doing so saves about 5 minutes of deploy time.  Let's see how this does in CI.  It seems to work locally when I provision/deprovision the same BMHs over and over again.


stuggi (approved):
/lgtm

PR #549, opened 2024-11-12
fix br slv undef

pinikomarov:
fix for fails on ocp install because delete_bridge_slaves.stderr is undef

### DIFF
--- a/ansible/ocp_ai_prep.yaml
+++ b/ansible/ocp_ai_prep.yaml
@@ -124,7 +124,6 @@
           type: bridge-slave
           state: absent
         register: delete_bridge_slaves
-        failed_when: delete_bridge_slaves.stderr != "" and "unknown connection" not in delete_bridge_slaves.stderr
         when: item != ""
         with_items:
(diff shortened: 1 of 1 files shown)


ansible/ocp_ai_prep.yaml
```diff
@@ -124,7 +124,6 @@
           type: bridge-slave
           state: absent
         register: delete_bridge_slaves
-        failed_when: delete_bridge_slaves.stderr != "" and "unknown connection" not in delete_bridge_slaves.stderr
```
olliewalsh, line 127: try:
```
failed_when: "unknown connection" not in delete_bridge_slaves.stderr|default("")
```

pinikomarov: that works thanks, changing

pinikomarov:
recheck

pinikomarov:
recheck

pinikomarov:
@olliewalsh @abays can we proceed with this ?

abays (approved):
/lgtm

PR #552, opened 2025-01-29
Add custom controller role for 17.1

stuggi:
Allows testing a custom controller role in 17.1 with fencing enabled.


abays (approved):
/lgtm

PR #554, opened 2025-02-24
[OSPK8-829] Expose interface to disable BM OSP compute cert validation

abays:
We need to be able to disable SSL cert validation for the BMCs for baremetal OSP computes.  This is needed for hybrid deployment scenario CI to get `idrac-redfish` to work properly, as we had to switch to this newer protocol after Metal3 deprecated and removed `idrac`.

Part of the work for this Jira: https://issues.redhat.com/browse/OSPK8-829


stuggi (approved):
/lgtm

PR #560, opened 2025-04-11
[OSPRH-15719] Improve resiliency of overcloud node reboot

abays:
The reboot logic for the "register overcloud nodes with subscription manager" playbook is somewhat vulnerable to intermittent network connectivity errors and slow node reboot times.  This has plagued CI runs recently and caused false failures where otherwise the job was likely to succeed.  This PR improves the playbook so that it is more tolerant of unstable networking and slow overcloud nodes.

Jira: https://issues.redhat.com/browse/OSPRH-15719

### DIFF
--- a/ansible/templates/osp/rhsm.yaml.j2
+++ b/ansible/templates/osp/rhsm.yaml.j2
@@ -1,8 +1,16 @@
 ---
 - hosts: overcloud
   become: true
+  gather_facts: false
 
   tasks:
+  - name: Wait for all overcloud nodes to be available
+    wait_for_connection:
+      connect_timeout: 10
+      sleep: 5
+      delay: 5
+      timeout: {{ ((default_timeout | int) * 6) if (ocp_bm_extra_workers | default ({}) | length) > 0 else ((default_timeout | int) * 3) }}
+
 {% if osp_registry_method == "rhsm" %}
   - name: Red Hat Subscription Management configuration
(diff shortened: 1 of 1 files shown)


ansible/templates/osp/rhsm.yaml.j2
```diff
@@ -1,8 +1,16 @@
 ---
 - hosts: overcloud
   become: true
+  gather_facts: false
```
abays, line 4: Gathering facts is not necessary for the playbook and in fact can introduce an avoidable error if a connection cannot be established with one or more nodes.  So we skip this and instead wait for all connections to be available via the new task on L7-12.

ansible/templates/osp/rhsm.yaml.j2
```diff
@@ -1,8 +1,16 @@
...
 
   tasks:
+  - name: Wait for all overcloud nodes to be available
+    wait_for_connection:
+      connect_timeout: 10
+      sleep: 5
+      delay: 5
+      timeout: {{ ((default_timeout | int) * 6) if (ocp_bm_extra_workers | default ({}) | length) > 0 else ((default_timeout | int) * 3) }}
```
abays, line 12: If real baremetal machines are used for computes, allow for extra time due to slow reboots

stuggi (approved):
/lgtm

PR #566, opened 2025-05-27
Forward the reverse zones for all network to IPA DNS when TLSe is enabled

olliewalsh:
PTR records must resolve to generate TLS certs for etcd at the edge sites.


stuggi (approved):
/lgtm

PR #568, opened 2025-07-02
Update masters disk size and add retries to system update

itzikb-redhat:
40G seems to be small in some installations.
On some setups the system update takes more time then others.


itzikb-redhat:
recheck

stuggi (approved):
/lgtm

PR #577, opened 2025-08-11
Collect overcloud system logs and add workaround for  Jira 18571

stuggi:
* Adds a playbook with make target to go collect a tar of the system logs of the overcloud nodes for debugging. Not collecting a full sosreport due to size limitation we have to store them
* also adds a workaround for Jira https://issues.redhat.com/browse/OSPRH-18571 , where on slow ci systems, the ovn bundle on the 16.2 base deployment might not be successful on all nodes, especially controller-0. This node is used on an ffu to bootstrap the raft cluster. If base deployment was note successful on 16.2 on this node, the raft cluster can have broken data. This workaround make sure to restart the bundle in case an error was recorded on pcs output after the 16.2 deployment
* moves 16.2_fencing, 16.2_fencing_nfs and 17.2_fencing, 17.2_fencing_nfs to a master/worker combo deployment to deduce the required resources

Jira: OSPRH-18571
Jira: OSPRH-15717

### DIFF
--- a/ansible/files/osp/16.2/extraconfig/post_deploy/validate_ovn_bundle.yaml
+++ b/ansible/files/osp/16.2/extraconfig/post_deploy/validate_ovn_bundle.yaml
@@ -1,0 +1,33 @@
+heat_template_version: 2015-04-30
+
+parameters:
+  servers:
+    type: json
+  EndpointMap:
+    default: {}
+    description: Mapping of service endpoint -> protocol.
+    type: json
+
+resources:
+  CustomPostConfigOvnCheck:
+    type: OS::Heat::SoftwareConfig
+    properties:
+      group: script
+      config: |
+        #!/bin/bash
+        # This script will only execute on the node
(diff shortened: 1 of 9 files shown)


ansible/files/osp/16.2/extraconfig/post_deploy/validate_ovn_bundle.yaml
```diff
@@ -0,0 +1,33 @@
...
+resources:
+  CustomPostConfigOvnCheck:
+    type: OS::Heat::SoftwareConfig
+    properties:
+      group: script
+      config: |
+        #!/bin/bash
+        # This script will only execute on the node
+        # with the short hostname containing 'controller-0'.
+        if [[ $(hostname -s) == *"controller-0"* ]]; then
```
olliewalsh, line 20: wonder if this could be more generic?

olliewalsh (approved):

PR #578, opened 2025-08-18
Make sure workaround exits with 0 when ovn bundle is fine

stuggi:
If grep can not find a failing ovn bundle the script exist with 1 which makes the post deployment workaround to fail.

Jira: OSPRH-18571


abays (approved):
/lgtm

PR #580, opened 2025-10-02
AI baremetal network config MAC address field name fix

abays:
Latest version of Assisted Installer API uses NMState format for certain network config, and we need to adjust our templates to align with that.


stuggi (approved):
/lgtm

PR #586, opened 2026-02-04
Add the csv to the 17.1_fencing.yaml for minor update job

stuggi:
Need to set a csv for the job to update to. lets default to the qe csv. if needed it can be customized via local-defaults.yaml.


abays (approved):
/lgtm

PR #591, opened 2026-07-17
Replace hardcoded FreeIPA and node root passwords with dynamic generation

abays:
Use Ansible's built-in password lookup to generate random credentials on first use and cache them in working_vars_dir for consistency across playbook runs. This eliminates well-known default passwords (abcd4321, dcba1234, 12345678) from the codebase. Users can still override via local-defaults.yaml.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


stuggi (approved):
/lgtm

PR #592, opened 2026-07-22
Add dependabot config and pin GitHub Actions to digest

stuggi:
Add dependabot.yml to automatically bump github-actions weekly. Pin actions/checkout and ansible/ansible-lint to SHA digests instead of floating tags for supply-chain security.


abays (approved):
/lgtm

PR #593, opened 2026-07-22
Remove hardcoded demo user with cluster-admin privileges

stuggi:
The demo playbook created an htpasswd user demo/demo with cluster-admin access. This could be a security risk and redundant since kubeadmin already provides cluster-admin login.

Removes demo.yaml, files/demo/, and the Makefile demo target.


abays (approved):
/lgtm

PR #601, opened 2026-08-03
Secure RPM downloads: use HTTPS for rhos-release

abays:
Switch rhos-release RPM download URL from plain HTTP to HTTPS in both the role defaults and the rhsm template.  GPG check bypasses are retained because these packages are not signed.


stuggi (approved):
/lgtm

